PR #14, opened 2019-06-08
[config] write default config for organisations #12

### LINKED ISSUE #12: [config] Decide role variables for configuration
@nemesisdesign

Vivekrajput20:
closes #12


org-configurations/default-configuration.yml
```diff
@@ -0,0 +1,88 @@
+---
+name: "default name"
+slug: "default-slug"
+
+# configuration variables for the server app
+server:
```
Vivekrajput20: @nemesisdesign I wrote this default config. I have separated variables in server and client section. Variables in the server section will only be available to the server and will prevent sensitive info like the secret key to be exposed to a client. The urls are highly customizable and organizations can set different host servers for proxy and also customize the url patterns for different views.  What do you think about it?

org-configurations/default-configuration.yml
```diff
@@ -0,0 +1,88 @@
...
+    - "it"
+  default_language: "en"
+  components:
+    header:
+      logo:
+        url: null # logo url
+        alternate_text: null
+      links:
+        - text:
+            en: "link1" #link text in english
```
Vivekrajput20: I've tried to make most of the things configurable in frontend. Orgs can provide texts in different languages like this.

org-configurations/default-configuration.yml
```diff
@@ -0,0 +1,88 @@
...
+            en: "link1"
+            it: null
+          url: "url1"
+        - text:
+            en: "link2"
+            it: null
+          url: "link2url"
+
+    login_form:
+      input_field:
```
Vivekrajput20: Orgs can add extra fields and we can generate HTML by looping through input_fields array.

org-configurations/default-configuration.yml
```diff
@@ -0,0 +1,88 @@
...
+          value:
+            en: 'Sign up for free' # text for register button
+            it: null
+      links:
+        - forget_password :
+            en: "Forgot your password?"
+            it: null
+        - account:
+            en: "Manage your account"
+            it: null
```
Vivekrajput20: @nemesisdesign  We can separate configuration for each component to a different YAML file and write an extensive default config for each component which can be overridden by organisations. Should I go ahead with such configs and write a config for other components as well? Or we don't require this much customization in our app?
Vivek

nemesisdesign: let's keep it simple for now, at max we can allow having one file for each organization if that does not complicate things too much

org-configurations/default-configuration.yml
nemesisdesign: very good!

org-configurations/default-configuration.yml
nemesisdesign: :+1:

PR #37, opened 2019-12-05
Allow custom ports #34

### LINKED ISSUE #34: [config] Dev-server dynamically assign port.
Currently, webpack dev server is configured to start at port 8080 and nodemon server at port 3030.
We need to configure it to accept a port number as a CLI parameter and if the parameter is not provided it should use the first available port starting from 8080 and 3030 respectively.

koenigscode:
Currently the webpack dev server and nodemon are running on a fixed
port.

This commit allows developers to choose their own port.
If no port was specified and the default one is already in use,
it will automatically find the next free port.

For information on how to set the ports, take a look at the readme.

This fixes #34

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -5,9 +5,27 @@ const HtmlWebpackPlugin = require("html-webpack-plugin");
 const path = require("path");
 
 const CURRENT_WORKING_DIR = process.cwd();
+const DEFAULT_PORT = 8080;
+const DEFAULT_SERVER_URL = "http://localhost:3030";
 
 module.exports = (env, argv) => {
-  const config = {
+  // Use user-specified port; if none was given, fall back to the default
+  // If the default port is already in use, webpack will automatically use
+  // the next available port
+  let port = argv.port || DEFAULT_PORT;
+
+  // The url the server is running on; if none was given, fall back to the default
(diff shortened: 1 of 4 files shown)


Vivekrajput20:
@koenigscode  I'm not able to run client on a custom port by running `npm run client --port 8085`
Also in you PR both nodemon and webpack is configured with same `port` arg, but when we'll `run npm run start` both the servers will be started simultaneously and we need two different arguments for defining ports for both the servers.

koenigscode:
Please take a look at the readme on how to start it with a custom port. you seem to have forgotten the -- before the --port argument.

as of starting both of them with npm start: I can implement that today; shouldn't be much work

koenigscode:
To support the npm run start command I had to do things a little differently (as I wanted a clean solution and not a workaround)

Please have a look at it again :)
The new way to start it is in the readme and works with environment variables. I added examples for Windows and Linux

atb00ker:
OpenWISP isn't really compatible with Windows, so I am not sure how those examples will help but I'll leave this review to @Vivekrajput20 and @nemesisdesign ! :smile:

Vivekrajput20:
Tested on Linux. Works fine. Good work @koenigscode . Can be approved

PR #38, opened 2019-12-07
[Fix #32] Finish the password change component

### LINKED ISSUE #32: [component] Design and develop password change component.
Openwisp-radius has a feature to allow users to change the password. The node proxy server already has [controller](https://github.com/openwisp/openwisp-wifi-login-pages/blob/master/server/controllers/password-change-controller.js) to change the password. However, the frontend app does not have a password change component because it was not there in the previous project.
Design and develop the react component to allow users to change the password from the status page.

someshkar:
Fixes #32 

Hey! This is v1 of the password change component, please have a look!

Here's what it looks like:

Screenshot from 2019-12-07 14-52-10

You can view it on the `/default/change-password` route.

Regards,
Somesh

### DIFF
--- a/client/components/organization-wrapper/organization-wrapper.js
+++ b/client/components/organization-wrapper/organization-wrapper.js
@@ -16,6 +16,7 @@ import PasswordConfirm from "../password-confirm";
 import PasswordReset from "../password-reset";
 import Registration from "../registration";
 import Status from "../status";
+import PasswordChange from "../password-change";
 
 export default class OrganizationWrapper extends React.Component {
   constructor(props) {
@@ -91,6 +92,14 @@ export default class OrganizationWrapper extends React.Component {
                 path={`${match.path}/status`}
                 render={() => {
                   if (isAuthenticated) return <Status cookies={cookies} />;
+                  return <Redirect to={`/${orgSlug}/login`} />;
+                }}
+              />
+              <Route
+                path={`${match.path}/change-password`}
+                render={() => {
+                  if (isAuthenticated)
+                    return <PasswordChange cookies={cookies} />;
                   return <Redirect to={`/${orgSlug}/login`} />;
                 }}
               />
--- a/client/components/password-change/index.css
+++ b/client/components/password-change/index.css
@@ -1,44 +1,52 @@
-.owisp-passchange-page {
-  max-width: 50%;
-  max-height: 100%;
+.owisp-password-change-container {
+  flex-grow: 1;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  align-items: flex-start;
+  width: 100%;
+  box-sizing: border-box;
+  padding: 27px 0 27px 0;
+  font-weight: bold;
+}
+
+.owisp-password-change-form {
+  max-width: 400px;
+  min-width: 400px;
+  background-color: #d3d3d3;
+  padding: 30px;
   display: flex;
   flex-direction: column;
-  align-items: top;
-  margin-left: 24%;
+  justify-content: flex-start;
 }
 
-.owisp-passchange-container {
-  max-width: 400px;
-  max-height: 100%;
-  margin-top: 30px;
-  background-color: #eee;
-  padding: 40px;
-  margin-bottom: 48.1vh;
-}
-
-.title {
+.owisp-password-change-form-title {
   font-size: 20px;
   font-weight: bold;
+  margin-top: 10px;
+  margin-bottom: 20px;
 }
 
-.owisp-passchange-input {
-  padding: 0 8px;
-  border-radius: 4px;
-  font-size: 14px;
+.owisp-password-change-input {
   height: 33px;
-  border: 1px solid #ccc;
+  line-height: 40px;
+  padding: 0 8px;
   box-sizing: border-box;
-  width: 21rem;
+  border: 1px solid #ccc;
+  border-radius: 4px;
+  box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075);
+  background-color: #fff;
+  color: #000;
   font-weight: bold;
-  margin-top: 0.5rem;
-  margin-bottom: 0.5rem;
+  font-size: 14px;
+  width: 100%;
 }
 
-.owisp-passchange-submit {
+.owisp-password-change-btn {
   display: block;
   margin-top: 10px;
   padding: 10px 20px;
-  background: #444;
+  background-color: rgba(0, 0, 0, 0.5);
   outline: none;
   border-radius: 5px;
   color: #fff;
@@ -48,36 +56,55 @@
   display: flex;
   justify-content: center;
   align-items: center;
+  border: none;
 }
 
-.owisp-passchange-submit:hover {
-  transition: all 0.3s ease-in;
-  transform: translateY(-1px);
-  background-color: #2b2b2b;
+.owisp-password-change-btn:hover {
+  background-color: rgba(0, 0, 0, 0.7);
 }
 
-.owisp-passchange-error {
-  color: red;
-  font-size: 12px;
-  font-weight: bold;
-}
+.owisp-password-change-error {
+  margin-top: 3px;
 
-.owisp-passchange-label {
   font-size: 14px;
-  font-weight: bold;
-  display: block;
-}
 
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  flex-wrap: nowrap;
+  align-items: center;
+  width: 100%;
+  color: #ffffff;
+  background: #d93025;
+  border-radius: 3px;
+  padding: 4px;
+  box-sizing: border-box;
+}
+.owisp-password-change-error-icon {
+  display: inline-block;
+  min-width: 16px;
+  text-align: center;
+  margin-right: 4px;
+  font-weight: 600;
+}
+.owisp-password-change-label {
+  width: 100%;
+  margin-bottom: 10px;
+}
+.owisp-password-change-label-text {
+  color: #000;
+  text-transform: capitalize;
+}
 @media (max-width: 700px) {
-  .owisp-passchange-page {
+  .owisp-password-change-container {
     max-width: 95%;
     margin-left: 20px;
   }
-  .owisp-passchange-container {
-    width: 80%;
+  .owisp-password-change-form {
+    min-width: 80%;
     padding: 25px;
   }
-  .owisp-passchange-input {
+  .owisp-password-change-input {
     width: 18rem;
   }
 }
--- a/client/components/password-change/index.css
+++ b/client/components/password-change/index.css
@@ -1,0 +1,83 @@
+.owisp-passchange-page {
+  max-width: 50%;
+  max-height: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: top;
+  margin-left: 24%;
+}
+
+.owisp-passchange-container {
+  max-width: 400px;
+  max-height: 100%;
+  margin-top: 30px;
+  background-color: #eee;
+  padding: 40px;
+  margin-bottom: 48.1vh;
+}
+
+.title {
+  font-size: 20px;
+  font-weight: bold;
+}
+
+.owisp-passchange-input {
+  padding: 0 8px;
+  border-radius: 4px;
+  font-size: 14px;
+  height: 33px;
+  border: 1px solid #ccc;
+  box-sizing: border-box;
+  width: 21rem;
+  font-weight: bold;
+  margin-top: 0.5rem;
+  margin-bottom: 0.5rem;
+}
+
+.owisp-passchange-submit {
+  display: block;
+  margin-top: 10px;
+  padding: 10px 20px;
+  background: #444;
+  outline: none;
+  border-radius: 5px;
+  color: #fff;
+  font-weight: bold;
+  cursor: pointer;
+  width: max-content;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.owisp-passchange-submit:hover {
+  transition: all 0.3s ease-in;
+  transform: translateY(-1px);
+  background-color: #2b2b2b;
+}
+
+.owisp-passchange-error {
+  color: red;
+  font-size: 12px;
+  font-weight: bold;
+}
+
+.owisp-passchange-label {
+  font-size: 14px;
+  font-weight: bold;
+  display: block;
+}
+
+@media (max-width: 700px) {
+  .owisp-passchange-page {
+    max-width: 95%;
+    margin-left: 20px;
+  }
+  .owisp-passchange-container {
+    width: 80%;
+    padding: 25px;
+  }
+  .owisp-passchange-input {
+    width: 18rem;
+  }
+}
--- a/client/components/password-change/index.js
+++ b/client/components/password-change/index.js
@@ -1,0 +1,16 @@
+import {connect} from "react-redux";
+
+import Component from "./password-change";
+
+const mapStateToProps = state => {
+  return {
+    passwordChange:
+      state.organization.configuration.components.password_change_page,
+    language: state.language,
+    orgSlug: state.organization.configuration.slug,
+  };
+};
+export default connect(
+  mapStateToProps,
+  null,
+)(Component);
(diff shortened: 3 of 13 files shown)


someshkar:
Hey! I squashed my commits into one, is that alright now?

Regards,
Somesh

atb00ker:
Hi @someshkar 

Look here: https://github.com/openwisp/openwisp-wifi-login-pages/pull/38/commits
There are 5 commits and the guidelines are not followed either. 
Hint: 
- You'll need to rebase
- You'll need to change the commit message and body
( Read the link I've send above )

If you are having problem, please ask in the the gitter channel: https://gitter.im/openwisp/general

client/components/organization-wrapper/organization-wrapper.js
```diff
@@ -90,8 +91,15 @@ export default class OrganizationWrapper extends React.Component {
               <Route
                 path={`${match.path}/status`}
                 render={() => {
-                  if (isAuthenticated) return <Status cookies={cookies} />;
-                  return <Redirect to={`/${orgSlug}/login`} />;
+                  // if (isAuthenticated) return <Status cookies={cookies} />;
```
Vivekrajput20: Why are these lines commented. Please undo this.

org-configurations/default-configuration.yml
```diff
@@ -4,7 +4,7 @@ slug: "default"
 
 # configuration variables for the server app
 server:
-  host: "http://127.0.0.1:8000"
+  host: "http://ec2-18-191-159-195.us-east-2.compute.amazonaws.com:8000"
```
Vivekrajput20: Revert all the changes made to this file.

client/components/password-change/password-change.js
```diff
@@ -0,0 +1,39 @@
+import "./index.css";
+import React, {useState} from "react";
+
+const PasswordChange = () => {
+  const [oldPassword, changeOldPassword] = useState();
+  const [newPassword, changeNewPassword] = useState();
+
+  const handleSubmit = (oldPassword, newPassword) =>
```
Vivekrajput20: Please implement this handleSubmit function. This will make an AJAX call to the node server, which will handle the password change and return some data. If the password is changed successfully  show success message, else show error message. Refer to [this](https://github.com/openwisp/openwisp-wifi-login-pages/blob/299460ed0b73cb9bc03821afed51667561861425/client/components/login/login.js#L50) to get an idea of how it needs to be done.

server/index.js
```diff
@@ -14,7 +14,9 @@ app.use(cookieParser());
 app.use(cookiesMiddleware());
 app.use(express.json());
 app.use(express.urlencoded({extended: false}));
+
```
Vivekrajput20: revert back changes made to this file please.

client/components/password-change/password-change.js
```diff
@@ -0,0 +1,39 @@
...
+        onClick={() => handleSubmit(oldPassword, newPassword)}
+        onKeyDown={() => handleSubmit(oldPassword, newPassword)}
+      >
+        Submit
+      </div>
+    </div>
+  );
+};
+
+export default PasswordChange;
```
Vivekrajput20: Also, I think there is no need to write this in a separate component. Include it to the PasswordChangePage component and rename that to PasswordChange.

client/components/password-change/password-change-page.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+
+import PasswordChange from "./password-change";
+
+const PasswordChangePage = () => {
+  return (
+    <div
+      style={{
```
Vivekrajput20: remove this inline style. Since we are using css files to style the components, keep it uniform and move these to css file.

client/components/password-change/index.css
```diff
@@ -0,0 +1,49 @@
+.owisp-passchange-container {
+  width: 30vw;
```
Vivekrajput20: Please avoid using vh and vw for now. See styles of any other component and follow the same structure and styles for the password change component.

client/components/password-change/password-change.js
someshkar: Right. About that, where do I `POST` the form data? And do I run some checks with something like [Joi](https://www.npmjs.com/package/@hapi/joi) before posting? If possible, could you point me to the file which handles the `POST` request on the server/proxy?

Vivekrajput20: @someshkar  [here](https://github.com/openwisp/openwisp-wifi-login-pages/blob/299460ed0b73cb9bc03821afed51667561861425/server/routes/account.js#L14) are the routes of the proxy node server. You have to make a post request to the password change route. No, there is no need to use Joi for now.

someshkar:
Hey @Vivekrajput20 !

I just finished all the changes you requested, could you have a look?

Thanks,
Somesh

org-configurations/default-configuration.yml
nemesisdesign: why are you changing this? It looks like something that should not be done to me

someshkar: Really sorry. I keep forgetting to change that back before committing. I've changed it back now though.

someshkar:
Hey @nemesisdesign !

Could you have a look now? The Travis build passed but I don't know why the coverage decreased.

Thanks a lot!
Somesh

someshkar:
@Vivekrajput20 I just added the `isAuthenticated` check.

client/components/password-change/index.js
```diff
@@ -0,0 +1,15 @@
+import {connect} from "react-redux";
+
+import Component from "./password-change";
+
+const mapStateToProps = state => {
+  return {
+    passwordConfirm: state.organization.configuration.components.confirm_form,
```
Vivekrajput20: @someshkar , `organization.configuration.components.confirm_form` variable you are using is declared [here](https://github.com/openwisp/openwisp-wifi-login-pages/blob/299460ed0b73cb9bc03821afed51667561861425/org-configurations/default-configuration.yml#L138). Check how settings for the password reset confirm form is defined inside that variable. Similary you have to declare a variable `password_change_from` and define all the settings for the password change component inside it. In this way, we can keep the settings configurable at organization level.

client/components/password-change/password-change.js
```diff
@@ -0,0 +1,75 @@
...
+        newPassword2,
+      }),
+    }).catch(error => {
+      console.log(error);
+    });
+  };
+
+  return (
+    <div className="owisp-passchange-container">
+      <div className="title">Change your password</div>
```
Vivekrajput20: @someshkar  Please do not hard code anything. Everything should be defined in the [default-config.yml](https://github.com/openwisp/openwisp-wifi-login-pages/blob/master/org-configurations/default-configuration.yml) and you should render fields, inputs, titles, placeholders, buttons etc accordingly. `Change your password` should come from the config file and should change to different language when the language is switched.

client/components/password-change/password-change.js
```diff
@@ -0,0 +1,75 @@
...
+        onClick={() => handleSubmit(newPassword1, newPassword2)}
+        onKeyDown={() => handleSubmit(newPassword1, newPassword2)}
+      >
+        Submit
+      </div>
+    </div>
+  );
+};
+
+const PasswordChangePage = ({orgSlug}) => {
```
Vivekrajput20: @someshkar  please merge both the components, there is no need to define 2 components here.

org-configurations/default-configuration.yml
```diff
@@ -192,9 +192,9 @@ client:
     status_page:
       content:
         en: |
-           WiFi Login Successful!
-           You can now use the internet.
-           You may leave this page open in case you want to log out.
+          WiFi Login Successful!
+          You can now use the internet.
+          You may leave this page open in case you want to log out.
```
Vivekrajput20: Revert these changes please.

nemesisdesign: I still see this request not being met

Vivekrajput20:
Still more work needs to be done @nemesisdesign.

client/components/organization-wrapper/organization-wrapper.js
```diff
@@ -16,6 +16,7 @@ import PasswordConfirm from "../password-confirm";
 import PasswordReset from "../password-reset";
 import Registration from "../registration";
 import Status from "../status";
+import PasswordChange from "../password-change/password-change";
```
Vivekrajput20: You're trying to import the component directly from the `password-change.js` file which is not connected to the redux state. The component is connected to the redux state in index.js file. Try importing from there.
Also, please follow the exact same practice that is being followed in the other components while writing the org-configuration file.

someshkar: Alright, I've fixed that now. Can you have a look?

someshkar:
@Vivekrajput20 I've fixed what you had asked for. Can you have a look?

org-configurations/default-configuration.yml
someshkar: Actually, I have prettier installed, and it tries to maintain uniformity in `.json` and `.yml` files. Shouldn't we keep this change since we're using double quotes everywhere in this file other than these few spots?

someshkar:
The Travis build seems to be failing because there's an issue with the `Status` component rendering. Should I update the snapshot?

org-configurations/default-configuration.yml
Vivekrajput20: @someshkar double quotes is fine, but you have deleted two lines in the content section of status page. Please revert that.

client/components/password-change/password-change.js
```diff
@@ -0,0 +1,109 @@
...
+    };
+
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleSubmit(e) {
+    if (e) e.preventDefault();
+    const {orgSlug} = this.props;
+    const url = passwordChangeApiUrl.replace("{orgSlug}", orgSlug);
+    const {newPassword1, newPassword2} = this.state;
```
Vivekrajput20: Please add a validation to confirm both the fields have same password before making a request. Something like [this](https://github.com/openwisp/openwisp-wifi-login-pages/blob/226d80e60c3200b4e05a77e7472c1b56c9b90bba/client/components/registration/registration.js#L40)

client/components/password-change/password-change.js
```diff
@@ -0,0 +1,109 @@
...
+import React from "react";
+import axios from "axios";
+import qs from "qs";
+import PropTypes from "prop-types";
+
+import {passwordChangeApiUrl} from "../../constants";
+
+import getText from "../../utils/get-text";
+
+// const PasswordChange = ({passwordChange, language, orgSlug}) => {
```
Vivekrajput20: Remove this.

Vivekrajput20:
@someshkar you have deleted a few lines from the default config of status page. That's why the test is failing. Revert that please. Also, you need to add a link to the password change page in the status page. Add the link and then update the snapshot for the status component.

someshkar:
@nemesisdesign are you running the `openwisp-radius` server on your port `8000` before starting the React app? It depends on the backend for things like the `orgSlug` and the Redux state for the props that the error you're having says are missing.

someshkar:
@Vivekrajput20 Do I add the link to the password change page from the status page in `default-configuration.yml` or should I hardcode it in the status component? I haven't updated the snapshot for now because of this.

Vivekrajput20:
@someshkar since all the organizations will need this link, we should add it to the default configuration only. Add something like this
links:
  -  text:
       -en: "change your password"
     url: "/{org_slug}/password-change/"
So that orgs can add multiple links if they want and if the link starts with / , parse it in the component and replace {org_slug} with the active slug.
What do you say @nemesisdesign ?

someshkar:
@Vivekrajput20 could you help me with the YAML syntax I need to use to implment this? Just have a look at the latest push. Thanks.

Vivekrajput20:
@someshkar  links can be added like [this](https://github.com/openwisp/openwisp-wifi-login-pages/blob/226d80e60c3200b4e05a77e7472c1b56c9b90bba/org-configurations/default-configuration.yml#L206). This will create an array of links and you can iterate over links to render them.Please don't add the icon key, we don't need icons for these links.

Vivekrajput20:
@someshkar  it's not working, I tried logging in but the status page goes blank once authentication is successful. There is some error log in console. Please fix the issue.

nemesisdesign:
> So that orgs can add multiple links if they want and if the link starts with / , parse it in the component and replace {org_slug} with the active slug.
> What do you say @nemesisdesign ?

What is the need of users being able to change that URL? I'm neutral to this. If you think is useful, we can do it, otherwise we can keep it simple and just add the link directly in the code.

someshkar:
@Vivekrajput20 I pmed you on gitter about the aws instance you had set up for me, could you have a look?

someshkar:
@Vivekrajput20 I fixed the Travis build and updated the snapshot for the status page, after fixing the links error.

client/components/password-change/password-change.js
```diff
@@ -0,0 +1,123 @@
...
+    const url = passwordChangeApiUrl.replace("{orgSlug}", orgSlug);
+    const {newPassword1, newPassword2} = this.state;
+
+    if (newPassword1 !== newPassword2) {
+      this.setState({
+        errors: {
+          newPassword2: passwordConfirmError,
+        },
+      });
+    }
```
Vivekrajput20: @someshkar you have added a check to validate if both the fields have same value and you're updating state but you are not returning the function. This will lead to further execution of the function and request will be made to the API. Please add a return statement to avoid making extra request when the data is invalid.

someshkar:
@Vivekrajput20 can you check the Travis build? The `<Link />` component converts to `<ForwardRef>` when I update the jest snapshot.

Vivekrajput20:
@someshkar it's working fine. But here are some of the issues that still needs to be addressed

- width of the change password button is small and it renders `Change Password` in 2 lines. Give `width: max-content;` instead of 60px.
- It's not mobile-responsive. Please see the styles of any other page and make it mobile responsive like other pages.
- You've missed the label field for the password fields. There are just to text fields in the page with no labels. Please add an extra label field to the password1 and password2 like 
```
password1:
    label:
        en: "new password"
password2:
    label:
        en: "confirm new password"
```
And show the labels in the component.
- ` Warning: Failed prop type: Invalid prop passwordChange.title of type object supplied to PasswordChange, expected string.` please fix this warning.
- We also need to redirect user to the status page once the password has been successfully changed and show the error message in case of an error.

client/components/password-change/password-change.js
```diff
@@ -0,0 +1,124 @@
...
+            className="owisp-passchange-input"
+            type="password"
+            name="newPassword2"
+            placeholder={getText(
+              passwordChange.input_fields.password2.placeholder,
+              language,
+            )}
+            onChange={e => this.handleChange(e)}
+          />
+          {errors.password2 ? errors.password2 : null}
```
Vivekrajput20: You have set error for `newPassword2` key and are checking for `password2` which is always false. Please fix this and wrap the error message in div and style it. [Here](https://github.com/openwisp/openwisp-wifi-login-pages/blob/3402215b3603982ac06d0963d9b5474763883869/client/components/registration/registration.js#L263) is an error message Div. You should make it similar to that one.

someshkar:
@Vivekrajput20 how should I fix the Travis build? The `<Link />` component becomes a `<ForwardRef />` during the jest test. Have a look at the Travis build.

Vivekrajput20:
@someshkar please update the snapshots, it will fix the Travis build.

someshkar:
@Vivekrajput20 I already have updated the jest snapshot, but `react-router-dom`'s `<Link to="" />` component becomes a `<ForwardRef />` after it's built. Have a look at the Travis build to see exactly what I mean. Should I switch back to `<a href="">`

someshkar:
@nemesisdesign I fixed the Travis build, is it ready to be merged?

Vivekrajput20:
@someshkar I still couldn't find [these](https://github.com/openwisp/openwisp-wifi-login-pages/pull/38#issuecomment-570781699) changes in the PR.
@nemesisdesign we can not merge it until these changes are made.

someshkar:
@Vivekrajput20 It was already mobile optimised, you can check the media queries. I've implemented the other features you had asked for.

Vivekrajput20:
@someshkar I made some changes that were necessary before merging. You can have a look at them.
Good work :+1:

nemesisdesign:
Thanks @someshkar @Vivekrajput20! :+1: :tada:

PR #44, opened 2020-01-03
[Enchancement] Switch from username to email in sign-up and login #41

### LINKED ISSUE #41: Switch from username to email in sign-up and login
When signing up, we can sign up users by putting their email as their username, at login they will have to use their email.

This usually results in better user experience.

nikitaermishin:
This is a PR for commit which switches `openwisp-wifi-login-pages` from username to email in sign-up and login by using email instead of username on frontend side of the application and sending email to server as a value of `username` (so it doesn't change backend part).
***
##### Changes:
 - Updated `login.js` :
   - Switched from username `input_field` to email `input_field`
   - Changed state (now it works with `email` & `password`)
 - Updated `registration.js` :
   - Changed state (now it works with `email`, `password1` and `password2`)
   - Removed username `input_field`
 - Updated `default-configuration.yml`:
   - Switched from `username` to `email` in `input_field`  of `login_form`
 - Updated tests (`login.test.js` and `registration.test.js`)
 - Updated snapshots (`login.test.js.snap` and `registration.test.js.snap`)
***
Fixes #41

### DIFF
--- a/client/components/login/login.js
+++ b/client/components/login/login.js
@@ -19,7 +19,7 @@ export default class Login extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      username: "",
+      email: "",
       password: "",
       errors: {},
     };
@@ -28,12 +28,12 @@ export default class Login extends React.Component {
   }
 
   componentDidMount() {
-    const username = getParameterByName("username");
+    const email = getParameterByName("email");
     const token = getParameterByName("token");
-    if (username && token) {
+    if (email && token) {
       this.setState(
         {
-          username,
+          email,
           password: token,
         },
         () => {
@@ -50,7 +50,7 @@ export default class Login extends React.Component {
   handleSubmit(event) {
     if (event) event.preventDefault();
     const {orgSlug, authenticate} = this.props;
-    const {username, password, errors} = this.state;
+    const {email, password, errors} = this.state;
     const url = loginApiUrl.replace("{orgSlug}", orgSlug);
     this.setState({
       errors: {},
@@ -62,7 +62,7 @@ export default class Login extends React.Component {
       },
       url,
       data: qs.stringify({
-        username,
+        "username": email,
         password,
       }),
     })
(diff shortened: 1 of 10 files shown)


org-configurations/default-configuration.yml
```diff
@@ -217,15 +217,15 @@ client:
             icon: "twitter.svg"
             url: "#"
       input_fields:
-        username:
-          type: "text"
-          pattern: '[a-zA-Z@.+\-_\d]{1,150}'
+        email:
+          type: "email"
+          pattern: null
```
atb00ker: I think there should be a pattern check for emails as well.
@nikitaermishin any specific reason why we are not doing pattern check like `<something>@<somthing>.<something>` ?

Vivekrajput20:
@nikitaermishin sign up is failing currently. I have updated the task description. Please go through it again to know how you can test your work.

org-configurations/default-configuration.yml
atb00ker: @nikitaermishin Ah, I didn't factor compatibility, please proceed whatever is best for the compatilibity! :smile:

nikitaermishin: @atb00ker So, the best for compatibility will be to use both of them (if `pattern` is not supported for some very old browsers (like `Chrome 5`) it will ignore `pattern` and only use `type="email"`). It will provide good compatibility and also good validation, so I think it is what we're looking for :smile:

#### Compatibility tables:
 - [pattern](https://caniuse.com/#feat=input-pattern)
 - [type="email"](https://caniuse.com/#feat=input-email-tel-url)

client/components/login/login.js
```diff
@@ -345,7 +345,7 @@ Login.propTypes = {
       links: PropTypes.arrayOf(PropTypes.object),
     }),
     input_fields: PropTypes.shape({
-      username: PropTypes.object,
+      email: PropTypes.object,
```
Vivekrajput20: @nikitaermishin can you please make it more detailed using PropTypes.shape(). Like
```
email : PropTypes.shape({
type: PropType.string.isRequired
...
})
```
In this way, we can make sure a required field is defined and supplied to the component. For input fields only `type` will be required and all others will be optional.

Vivekrajput20:
Very good work @nikitaermishin . Everything looks good and works fine. I've suggested a minor improvement. 
@nemesisdesign I've tested the PR and it works fine. You can test it and then we can merge this.

PR #71, opened 2020-06-07
[npm] Add npm command to generate organization configuration file usi…

Vivekrajput20:
Added npm command `$ npm run add-org`.
This command currently asks user to enter the following values in prompt:
- name
- slug
- UUID
- secret key

and generates the configuration.yml file based on given inputs. Currently I've added these 4 required fields only, but this can be easily extended in future to add more required/optional fields like 'Do you want to use default assets?', 'host url' etc.


Vivekrajput20:
@nemesisdesign  please review it and let me know how can I improve it to make it easy for org admins to add the configurations.

readme.md
```diff
@@ -59,13 +59,15 @@ yarn upgrade
...
-List of variables required in organization configuration:
-
-- name
-- slug
-- uuid: uuid of the organization
-- secret_key: token of the organization
+##### Add Organization configuration
+run `$ npm run add-org`
+When you run this command you’re prompted to provide the following properties:
+| Property          | Description                                       |
```
nemesisdesign, line 65: leave a blank line here

readme.md
```diff
@@ -59,13 +59,15 @@ yarn upgrade
...
 
-Write the configuration of the organization in a yml file in `org-configuration` directory.
-List of variables required in organization configuration:
-
-- name
-- slug
-- uuid: uuid of the organization
-- secret_key: token of the organization
+##### Add Organization configuration
+run `$ npm run add-org`
```
nemesisdesign, line 63: run > Run
I'd add a semicolon at the end of the line. Check the result on github or in a preview.

Vivekrajput20: Run `$ npm run add-org`;
@nemesisdesign  it just adds the ; at the end of the line, is that what you want?

Vivekrajput20:
@nemesisdesign  I have changed secret_key and uuid to input fields. Currently organization token is
also used for signing the auth-token cookie, but we can shift to using uuid for signing it and then we can completely remove the dependency on secret key

Vivekrajput20:
@nemesisdesign  I've updated the PR adding a prompt "Do you want to copy the default assets for your organization?" 
If yes, default assets will be copied to the organization's asset directory else all those assets links are set to null in the configuration file so that there is no broken links on the page.
Please test it.

nemesisdesign:
> Currently organization token is
> also used for signing the auth-token cookie, but we can shift to using uuid for signing it and then we can completely remove the dependency on secret key

That will be great. We can get there later when we work also on removing the usage of the authorize API endpoint of openwisp-radius.

PR #72, opened 2020-06-21
[UI] Add loading indicator

### LINKED ISSUE #64: [wifi-login-pags] Add loading indicator
Add loading indicator that should be used to block the page during all form submissions.

It could be like the one used in http://openwisp.org/activity.html.

Vivekrajput20:
Closes #64

### DIFF
--- a/client/components/login/login.js
+++ b/client/components/login/login.js
@@ -5,21 +5,23 @@ import axios from "axios";
 import PropTypes from "prop-types";
 import qs from "qs";
 import React from "react";
-import {Link, Route} from "react-router-dom";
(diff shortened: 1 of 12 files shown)


Vivekrajput20:
@nemesisdesign  yes we have done that
[this](https://github.com/openwisp/openwisp-wifi-login-pages/blob/e80d6054394cf9ca005cf0ac004e461f7098fdab/server/controllers/registration-controller.js#L57)
is where we do that. But it checks for the returned response's status code and only logs 500. In this case there was not server running, so no returned response code and that's why it did not log it.

PR #76, opened 2020-07-25
[change] Switch to Bearer token

nemesisdesign:
This adds support for https://github.com/openwisp/openwisp-radius/pull/139, @Vivekrajput20 do you know if there are other parts that need to be changed?


Vivekrajput20:
@nemesisdesign  it's used in this component only.

Issue #79, opened 2020-08-23
[wifi-login-pages] Bug: fonts get blocked by captive portal


nemesisdesign:
We should not load fonts from the google fonts API: https://fonts.googleapis.com/css?family=Montserrat:500,600&display=swap.

This request gets blocked by the captive portal and fails.
Fonts should be shipped from the static files of the application.

iamhitman21:
May I work on this issue?

nemesisdesign:
@iamhitman21 go for it

PR #85, opened 2020-10-03
[bug] Fixed  blocking font in captive portal #79

iamhitman21:
Configured file-loader withwebpack, added fonts to the application and changed index.css to remove blocking font #79


Vivekrajput20:
Thanks!

PR #95, opened 2020-10-21
[feature] Allow specifying if links should be shown to authenticated/unauthenticated users only #93

### LINKED ISSUE #93: [menu] Allow specifying if links should be shown to authenticated or unauthenticated users
We should be able to specify whether links should be shown to any user, only authenticated users or unauthenticated users, eg:

```
links:
  - text:
      en: "Sign In" #link text in english
    url: "/<slug>/login"
    authenticated: false
  - text:
      en: "Sign Up"
    url: "/<slug>/registration"
    authenticated: false
  - text:
      en: "Change password"
    url: "/<slug>/change-password"
    authenticated: true
  - text:
      en: "About this service"
    url: "https://somewebsite.com/about"
```

In the example above, the first two links must be shown only to users who are not authenticated, the third link should be shown only to users which are authenticated and the last one to any user (current behavior).

NoumbissiValere:
Implements and closes #93

### DIFF
--- a/client/components/contact-box/contact.js
+++ b/client/components/contact-box/contact.js
@@ -6,10 +6,11 @@ import React from "react";
 
 import getAssetPath from "../../utils/get-asset-path";
 import getText from "../../utils/get-text";
+import shouldLinkBeShown from "../../utils/should-link-be-shown";
 
 export default class Contact extends React.Component {
   render() {
-    const {contactPage, language, orgSlug} = this.props;
+    const {contactPage, language, orgSlug, isAuthenticated} = this.props;
     const {email, helpdesk, social_links} = contactPage;
     return (
       <>
@@ -39,27 +40,30 @@ export default class Contact extends React.Component {
             </div>
             <div className="owisp-contact-links">
               {social_links.map(link => {
-                return (
-                  <a
-                    href={link.url}
-                    target="_blank"
-                    rel="noopener noreferrer"
-                    key={link.url}
-                    className={`owisp-contact-${getText(
-                      link.alt,
-                      language,
-                    )}-link owisp-contact-link`}
-                  >
-                    <img
-                      src={getAssetPath(orgSlug, link.icon)}
-                      alt={getText(link.alt, language)}
+                if (shouldLinkBeShown(link, isAuthenticated)) {
+                  return (
+                    <a
+                      href={link.url}
+                      target="_blank"
+                      rel="noopener noreferrer"
+                      key={link.url}
                       className={`owisp-contact-${getText(
                         link.alt,
                         language,
-                      )}-image owisp-contact-image`}
-                    />
-                  </a>
-                );
+                      )}-link owisp-contact-link`}
+                    >
+                      <img
+                        src={getAssetPath(orgSlug, link.icon)}
+                        alt={getText(link.alt, language)}
+                        className={`owisp-contact-${getText(
+                          link.alt,
+                          language,
+                        )}-image owisp-contact-image`}
+                      />
+                    </a>
+                  );
+                }
+                return null;
               })}
             </div>
             <div className="owisp-contact-inner"/>
@@ -70,6 +74,9 @@ export default class Contact extends React.Component {
   }
 }
 
+Contact.defaultProps = {
+  isAuthenticated: false
+};
 Contact.propTypes = {
   language: PropTypes.string.isRequired,
   orgSlug: PropTypes.string.isRequired,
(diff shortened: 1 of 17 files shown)


nemesisdesign (commented):
Haven't tested it yet but I already have one question, could you add a brief section in the README that explains how to use this?

client/utils/log-error.js
```diff
@@ -1,10 +1,14 @@
 const logError = (error, errorText = "") => {
-  console.error(
-    'Status',
-    error.response.status,
-    error.response.statusText,
-    ':',
-    errorText
-  );
+  try {
```
NoumbissiValere, line 2: added this try-catch block because in the latest version of node (v15.0.1),  there were instances where `error` didn't have the attribute `response` causing it to raise the error complained by `Travis buddy` above.

client/components/contact-box/contact.js
```diff
@@ -39,27 +39,30 @@ export default class Contact extends React.Component {
...
-                    key={link.url}
-                    className={`owisp-contact-${getText(
-                      link.alt,
-                      language,
-                    )}-link owisp-contact-link`}
-                  >
-                    <img
-                      src={getAssetPath(orgSlug, link.icon)}
-                      alt={getText(link.alt, language)}
+                if (link.authenticated === undefined || link.authenticated === isAuthenticated) {
```
nemesisdesign, line 42: this if looks repeated across different components, may be worth to move it to a simple function stored in utils

readme.md
```diff
@@ -138,6 +139,34 @@ Note that you need to tell the client the server's port
 (unless you're using the default server port, which is 3030)
 so the client knows where he can find the server.
 
+### Settings
+
+#### authenticated
```
nemesisdesign, line 144: I would change this to something like "Menu items".

readme.md
```diff
@@ -138,6 +139,34 @@ Note that you need to tell the client the server's port
...
+
+This setting is used to specify which link(s) should be visible to all users
+or only to authenticated users or to unauthenticated users.
+Let's consider the `Footer` configuration below;
```
nemesisdesign, line 148: I'd change this description to be generic, since no configuration setting was explained, we should try to explain to users what these settings are about: specify links in 3 different components and specify if they should be shown to any user, only authenticated users or only unauthenticated users.

readme.md
```diff
@@ -138,6 +139,34 @@ Note that you need to tell the client the server's port
...
+        en: "link-2"
+        url: "https://link-2.com"
+        authenticated: false
+    - text:
+        en: "link-3"
+        url: "https://link-3.com"
+        authenticated: true
+```
+
+with the settings above, link-1 will be visible to all users while link-3 visible
```
nemesisdesign, line 166: First letter should be capital letter.

readme.md
```diff
@@ -138,6 +139,34 @@ Note that you need to tell the client the server's port
...
+        en: "link-1"
+        url: "https://link-1.com"
+    - text:
+        en: "link-2"
+        url: "https://link-2.com"
+        authenticated: false
+    - text:
+        en: "link-3"
+        url: "https://link-3.com"
+        authenticated: true
```
nemesisdesign, line 163: I believe the indentation is wrong here.

```
- text:
    en: "link-1"
  url: "https://link-1.com"
```

client/components/footer/footer.test.js
```diff
@@ -7,6 +7,29 @@ import getConfig from "../../utils/get-config";
 import Footer from "./footer";
 
 const defaultConfig = getConfig("default");
+const footerLinks = [
+  {
+    text: {en: "link-1"},
+    url: "https://link-1.com",
+    authenticated: true,
```
nemesisdesign, line 14: would be good to use real world examples like the one I suggested for the docs also in the tests, it makes the tests more readable and the code is less costly to maintain as a result

client/components/header/header.js
```diff
@@ -163,32 +166,35 @@ export default class Header extends React.Component {
               } owisp-header-mobile-menu`}
           >
             {links.map((link, index) => {
-              if (isInternalLink(link.url)) {
+              if (link.authenticated === undefined || link.authenticated === isAuthenticated) {
+                if (isInternalLink(link.url)) {
```
nemesisdesign, line 170: I would avoid the nested `if`, eg:

```javascript
if(shouldLinkBeShown(link) && isInternalLink(link.url)) {
  // code
}
```

nemesisdesign (changes requested):
Good job Noumbissi, the solution works :+1: 

The implementation and docs need a bit of improvement but no major change, only adjustments and refinements, see below.

client/components/header/header.js
NoumbissiValere: Understanding the way this code was written, I think the second if should stay in place.

nemesisdesign (approved):

PR #105, opened 2020-11-13
[feature] Added option to specify host url #102

### LINKED ISSUE #102: [wifi-login-app] Add openwisp-radius URL parameter to add-org configuration command
The add-org configuration generator is missing a very important parameter, the openwisp-radius URL is currently not asked to the user. We should ask for this information, making it clear that it will default to `http://127.0.0.1:8000` if left empty.

The new doc section added in #101 needs to be updated to mention this new option.

NoumbissiValere:
Closes #102

### DIFF
--- a/internals/generators/organization/index.js
+++ b/internals/generators/organization/index.js
@@ -78,6 +78,12 @@ module.exports = {
       },
     },
     {
+      type: "input",
+      name: "openwisp_radius_url",
+      message: "What is the URL of openwisp-radius?",
+      default: "http://127.0.0.1:8000",
+    },
+    {
       type: "confirm",
       name: "assets_confirm",
       message: "Do you want to copy the default assets for your organization?",
--- a/internals/generators/organization/config.yml.hbs
+++ b/internals/generators/organization/config.yml.hbs
@@ -6,8 +6,8 @@ settings:
 
 # configuration variables for the server app
 server:
-  # put the URL of openwisp-radius here
-  host: "http://127.0.0.1:8000"
+  # URL to openwisp-radius
+  host: "{{openwisp_radius_url}}"
   proxy_urls:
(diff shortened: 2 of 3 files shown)


client/components/mobile-phone-verification/mobile-phone-verification.js
```diff
@@ -141,51 +186,6 @@ export default class MobilePhoneVerification extends React.Component {
...
-        });
-      });
-  }
-
-  async handleLogout() {
-    const {orgSlug, logout, cookies} = this.props;
-    logout(cookies, orgSlug);
-    toast.success(logoutSuccess);
-  };
-
```
nemesisdesign, line 188: why are you changing this?

nemesisdesign (changes requested):
Looks good, will test it soon. Please rebase on the current master, avoid changing other parts of the code in this PR.

internals/generators/organization/index.js
```diff
@@ -77,6 +77,12 @@ module.exports = {
         return "The captive portal logout action url required";
       },
     },
+    {
+      type: "input",
+      name: "host",
```
nemesisdesign, line 82: I would call it openwisp_radius_url rather than `host`, to make it more obvious.

nemesisdesign (changes requested):

client/components/mobile-phone-verification/mobile-phone-verification.js
nemesisdesign: I see, it shouldn't happen on the latest master, if it does, let me know and I'll fix master (but the build is passing so it should not happen).

NoumbissiValere: ok

nemesisdesign (approved):

PR #118, opened 2020-11-19
[feature] Active users can change phone number #106

### LINKED ISSUE #106: [wifi-login-pages] Implement change phone number for active users
To be done after https://github.com/openwisp/openwisp-radius/issues/179 is resolved.

This app has a change phone number component which is meant for inactive users to change their phone number, in case they mistakenly used a wrong phone number during registration.

- [ ] we have to change the component so it allows verified users to use it, right now active users get redirected to status
- [ ] add a test for the functionality described in the previous point
- [ ] we have to ensure the whole flow works:
  - change phone number generates a new SMS token, flags user as inactive and user is redirected to mobile verification component
  - once user confirms the SMS token, it's flagged as active again, their phone number is correctly updated and is redirected to status component 
- [ ] change the org config generator so that if mobile_phone_verification is enabled, a link to change phone number is added to the header links (shown only to authenticated users)

NoumbissiValere:
Closes #106

### DIFF
--- a/client/components/mobile-phone-change/mobile-phone-change.js
+++ b/client/components/mobile-phone-change/mobile-phone-change.js
@@ -119,9 +119,9 @@ class MobilePhoneChange extends React.Component {
           '"response_code" !== "AUTH_TOKEN_VALIDATION_SUCCESSFUL"',
         );
       } else {
-        const {phone_number, is_active} = response.data;
+        const {phone_number} = response.data;
         this.setState({phone_number});
-        verifyMobileNumber(!is_active);
+        verifyMobileNumber(true);
       }
       return true;
     } catch (error) {
@@ -140,14 +140,13 @@ class MobilePhoneChange extends React.Component {
       orgSlug,
       language,
       phone_number_change,
-      needsMobilePhoneVerification,
       settings
     } = this.props;
     const { input_fields } = phone_number_change;
     const { buttons } = phone_number_change;
 
     // check equality to false, it may be undefined
-    if (needsMobilePhoneVerification === false || !settings.mobile_phone_verification) {
+    if (!settings.mobile_phone_verification) {
       return <Redirect to={`/${orgSlug}/status`} />;
     }
 
@@ -258,8 +257,4 @@ MobilePhoneChange.propTypes = {
   cookies: PropTypes.instanceOf(Cookies).isRequired,
   logout: PropTypes.func.isRequired,
   verifyMobileNumber: PropTypes.func.isRequired,
-  needsMobilePhoneVerification: PropTypes.bool
-};
-MobilePhoneChange.defaultProps = {
-  needsMobilePhoneVerification: null
 };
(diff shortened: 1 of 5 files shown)


client/components/mobile-phone-change/mobile-phone-change.js
```diff
@@ -119,9 +119,9 @@ class MobilePhoneChange extends React.Component {
           '"response_code" !== "AUTH_TOKEN_VALIDATION_SUCCESSFUL"',
         );
       } else {
-        const {phone_number, is_active} = response.data;
+        const {phone_number} = response.data;
         this.setState({phone_number});
-        verifyMobileNumber(!is_active);
+        verifyMobileNumber(settings.mobile_phone_verification);
```
nemesisdesign, line 124: `settings.mobile_phone_verification` is checked elsewhere, just do `verifyMobileNumber(true)`, and add a preceding comment like "Changing the phone number always requires verification".

client/components/mobile-phone-change/mobile-phone-change.js
```diff
@@ -140,14 +140,13 @@ class MobilePhoneChange extends React.Component {
...
       language,
       phone_number_change,
-      needsMobilePhoneVerification,
       settings
     } = this.props;
     const { input_fields } = phone_number_change;
     const { buttons } = phone_number_change;
 
     // check equality to false, it may be undefined
-    if (needsMobilePhoneVerification === false || !settings.mobile_phone_verification) {
```
nemesisdesign, line 150: can you explain why this can be removed? Just to make sure I'm following you.

client/components/mobile-phone-change/mobile-phone-change.test.js
```diff
@@ -305,4 +324,12 @@ describe("Change Phone Number: corner cases", () => {
...
     expect(wrapper.find(Redirect)).toHaveLength(1);
   });
+
+  it("shouldn't redirect if user is active and mobile verificaton is true", async () => {
+    shouldRedirectActiveUserHelper(true);
+  });
+
+  it("should redirect if user is active and mobile verification is false", async () => {
+    shouldRedirectActiveUserHelper(false);
+  });
```
nemesisdesign, line 334: just assume mobile verification is true, no need to test when it's false because the component will not be rendered.
So please simplify the test code.

nemesisdesign (changes requested):
Good progress, tested and it works, see my comments below.
See also https://github.com/openwisp/openwisp-wifi-login-pages/issues/126 which needs to be addressed after this.

client/components/mobile-phone-change/mobile-phone-change.js
NoumbissiValere: This component can only be accessed if the user is authenticated. So it shouldn't be accessible only if `mobile_phone_verification` is true. 

- So if user is authenticated and mobile_phone_verification is true, user should be able to access this component and use it.
- if user is authenticated and mobile_phone_verification is false, user should not be able to access this component. hence he should be redirected

nemesisdesign (approved):
Thanks :+1: 

Please move on with https://github.com/openwisp/openwisp-radius/pull/183#pullrequestreview-540086320 so we can merge this.

PR #139, opened 2020-12-07
[feature] Added user info to status page #104

### LINKED ISSUE #104: [wifi-login-pages] Feature: show user details in status page
In the status page, we need to show the details of the currently logged in user:

- username (only if different than email and phone number)
- email
- phone number (if mobile phone verification enabled)

NoumbissiValere:
Closes #104

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,6 +21,7 @@
     "node": true
   },
   "rules": {
+    "jsx-a11y/label-has-associated-control": 0,
     "array-bracket-spacing": [2, "never"],
     "eol-last": ["error", "always"],
     "semi": [2, "always"],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,6 @@
     "node": true
   },
   "rules": {
-    "jsx-a11y/label-has-associated-control": 0,
     "array-bracket-spacing": [2, "never"],
     "eol-last": ["error", "always"],
     "semi": [2, "always"],
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -191,6 +191,7 @@ export default class Status extends React.Component {
       } else {
         const {radius_user_token: password, username, email, phone_number, is_active} = response.data;
         const userInfo = {};
+        userInfo.status = "";
         userInfo.email = email;
         if (username !== email && username !== phone_number) {
           userInfo.username = username;
@@ -223,6 +224,7 @@ export default class Status extends React.Component {
     const {content, links, buttons, user_info} = statusPage;
     const {username, password, sessions, userInfo} = this.state;
     const contentArr = getText(content, language).split("\n");
+    userInfo.status = getText(user_info.status.value, language);
     return (
       <>
         <div className="container content" id="status">
@@ -237,19 +239,14 @@ export default class Status extends React.Component {
                     );
                   return null;
                 })}
-              <br />
-              <table>
-                <tbody>
-                  {Object.keys(userInfo).map(key => {
+                {Object.keys(userInfo).map(key => {
                     return (
-                      <tr key={key}>
-                        <td>{getText(user_info[key].text, language)}:</td>
-                        <td>{userInfo[key]}</td>
-                      </tr>
+                      <p key={key}>
+                        <label>{getText(user_info[key].text, language)}:</label>
+                        <span>{userInfo[key]}</span>
+                      </p>
                     );
-                  })}
-                </tbody>
-              </table>
+                })}
 
               {links && (
                 <div className="links row">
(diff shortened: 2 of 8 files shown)


NoumbissiValere:
@nemesisdesign the complain is not part of this solution but I don't get it why linting is complaining in [phone-token-controller](https://github.com/openwisp/openwisp-wifi-login-pages/blob/master/server/controllers/mobile-phone-token-controller.js#L23) and not in [phone change number](https://github.com/openwisp/openwisp-wifi-login-pages/blob/master/server/controllers/mobile-phone-number-change-controller.js#L24) for the same thing.

NoumbissiValere:
[PR 202](https://github.com/openwisp/openwisp-radius/pull/202) is needed to test this work

client/components/status/status.js
```diff
@@ -188,8 +199,17 @@ export default class Status extends React.Component {
           '"response_code" !== "AUTH_TOKEN_VALIDATION_SUCCESSFUL"',
         );
       } else {
-        const {radius_user_token: password, username, is_active} = response.data;
-        this.setState({username, password, is_active});
+        const {radius_user_token: password, username, email, phone_number, is_active} = response.data;
+        const user_info = {};
```
nemesisdesign, line 203: use camel case in javascript code please

client/components/status/status.js
```diff
@@ -211,23 +231,25 @@ export default class Status extends React.Component {
...
-              })}
+              <table>
+                <tbody>
+                  {Object.entries(user_info).map(([key, value]) => {
+                      return (
+                        <tr key={key}>
+                          <td>{key}:</td>
+                          <td>{value}</td>
+                        </tr>
+                      );
```
nemesisdesign, line 249: Please define the fields one by one and get the translatable text in a way that is consistent with the rest of the application (create one config key for each word).

nemesisdesign (changes requested):
Good progress @NoumbissiValere, here's a quick review, will need to look more in depth into it tomorrow and test it as well.

nemesisdesign (changes requested):
- The status row indicated in the wireframe is missing.
- I would switch from using a table to using a series of `<p>` elements, `<label>` and span, eg:
`<p><label>Email:</label><span>email value</span></p>`
- I also suggest using the CSS of the component to style the label to be bold and to have some margin on the right side.
- Please when you need to add space between elements use CSS, not <br> elements, the space below is huge and looks bad:

Screenshot from 2020-12-09 21-41-20

NoumbissiValere:
@nemesisdesign should I remove this text:
```WiFi Login Successful!

You can now use the internet.

You may leave this page open in case you want to log out.
```

nemesisdesign:
@NoumbissiValere leave it for now, I will think about it a bit more.

.eslintrc.json
```diff
@@ -21,6 +21,7 @@
     "node": true
   },
   "rules": {
+    "jsx-a11y/label-has-associated-control": 0,
```
nemesisdesign, line 24: adding this rule here affects the whole project. It should be possible to add an ignore rule only in the interested file, which is a lot better.

client/components/status/status.js
```diff
@@ -211,23 +221,32 @@ export default class Status extends React.Component {
...
-                return null;
-              })}
+                  if (text !== "")
+                    return (
+                      <p key={text}>
+                        {text}
+                      </p>
+                    );
+                  return null;
+                })}
```
nemesisdesign, line 241: why did you increase the indentation here? It looks like an unnecessary change and I think it should be reverted.

internals/generators/organization/config.yml.hbs
```diff
@@ -281,6 +281,23 @@ client:
...
+          text:
+            en: "Email"
+        username:
+          text:
+            en: "Username"
+      {{#if mobile_phone_verification}}
+        phone_number:
+          text:
+            en: "Phone number"
+      {{/if}}
```
nemesisdesign, line 300: let's not use the if here, the reason is that if mobile phone verification is enabled at a later step, the user will have to modify manually the config file which is time consuming and error prone.

org-configurations/default-configuration.yml
```diff
@@ -208,6 +208,18 @@ client:
...
+          text:
+            en: "Status"
+          value:
+            en: "Logged in"
+        email:
+          text:
+            en: "Email"
+        username:
+          text:
+            en: "Username"
```
nemesisdesign, line 222: please add the mobile phone field here as well for consistency with the change requested above

nemesisdesign (changes requested):
Tested and looks good, please address the following concerns below so we can merge.

nemesisdesign (approved):

PR #141, opened 2020-12-10
[feature] Added active and inactive sessions to status #80

### LINKED ISSUE #80: [wifi-login-pages] Show active sessions + logout button for each session
It's time to enrich the status page to provide more information to the user about their session.

We can get all the active sessions of the user from openwisp-radius and list them.
Each session can be the row of a table.
If logging out by session is enabled (we need to make this configurable because not all captive portals can do this), we can show a logout on each session, this way the user can decide which of their device to log out, very useful when the user is allowed to use multiple devices to surf the internet.

The information of the active sessions should be reloaded every minute.

Ideally we should implement this together with #81, we should strive to make only one HTTP request to get both active and past sessions.

### LINKED ISSUE #81: [wifi-login-pages] Show past sessions (nodejs + openwisp-radius API)
In the status page, show past session of the user with support for pagination **(*)**, show the first 10 sessions by default.

Ideally we should implement this together with #80, we should strive to make only one HTTP request to get both active and past sessions.

### (*) Regarding pagination

We can use an infinite scroller component available on npm, eg: https://www.npmjs.com/package/react-infinite-scroller or https://www.npmjs.com/package/react-infinite-scroll-component, or also another one if we can find any package we think is better.

Let's choose the best compromise between stability and smallest size
I tried measuring the size of these two, there doesn't seem to be a lot of difference once the JS is gzipped:

```
package-size react-infinite-scroller react-infinite-scroll-component

  package                                size      minified  gzipped
  react-infinite-scroller@1.2.4          51.15 KB  6.82 KB   2.13 KB
  react-infinite-scroll-component@5.1.0  21.02 KB  7.68 KB   2.43 KB
```

The most popular one seems to be also smaller, so if that works, we could just go with that one.

NoumbissiValere:
Closes #81
Closes #80


internals/generators/organization/config.yml.hbs
```diff
@@ -410,6 +435,7 @@ client:
       fields:
         id: logout_id
       additional_fields: []
+      logout_by_session: true
```
nemesisdesign, line 438: can you please add a new prompt for this?
Something like: "Does your captive portal support log out by session ID?" yes/no (default to no).

nemesisdesign (changes requested):
It doesn't compile to me, I get the same failure of the Travis build:

```
✖ ｢wdm｣: 
[1] ERROR in ./components/status/status.js
[1] Module build failed (from ../node_modules/babel-loader/lib/index.js):
[1] SyntaxError: /home/nemesis/Code/openwisp/openwisp-wifi-login-pages/client/components/status/status.js: Can not use keyword 'await' outside an async function (269:4)
[1] 
[1]   267 |       this.logoutFormRef.current.submit();
[1]   268 |     }
[1] > 269 |     await this.getUserPassedRadiusSessions();
[1]       |     ^
[1]   270 |     await this.getUserActiveRadiusSessions();
[1]   271 |   }
[1]   272 | 
```

nemesisdesign (changes requested):
Great progress @NoumbissiValere! :+1: 

Please could your rebase it on the current master and fix conflicts?

I have tested it without looking much at the code yet, but I have a few requests for improvements: let's try to improve it further so is up to our basic UX quality standards. 

- there's a bit too much blank space between the upper boxes and the session box:
Screenshot from 2020-12-15 21-22-14

- if there's no active session like in the screenshot above, there's a blank space on the right, which is bad. I remember having asked you to separate the active sessions from the historic sessions, if you do that, we'll have then two tables:
  1. one table for the active sessions, only if active sessions are present, with the space for the logout button (if the setting is enabled)
  2. the second table with the historic sessions without the blank space on the right

- there's no kind of visual feedback when new sessions are being loaded, in these cases a loading indicator should be shown in some way somewhere (as we did in the notifications module), maybe we can add a specific indicator at the bottom of the page and configure the scroller so that new sessions are loaded only when the bottom of the page is visible (so the indicator would be shown while new sessions are loading), can you check the library docs to see how to do this? There must be a way to do it since it's a very common thing to do.

What do you think?

nemesisdesign (changes requested):
Great progress!

I found out another issue:

Screenshot from 2020-12-17 23-28-10

Try opening the page when there are no active sessions please and ensure it looks good.

Also try different window sizes:

Screenshot from 2020-12-17 23-33-11

Setting width to 100% to all elements involved should fix it.

On narrow screens it looks pretty much unreadable, take a look:

Screenshot from 2020-12-17 23-35-16

Many users will use these services from mobile phones so it's a shame they'd not be really able to use this, right?
Can you please try [this CSS technique for narrow tables](https://css-tricks.com/responsive-data-tables/) (or anything similar/better you can find)?

responsive-table-on-iphone-1

In the technique described at this link, every row becomes a section in which every header is repeated on the left. Since the headers on the left are created with CSS, maybe you can use the JSS (the CSS written in JS) in the component itself, so you should be able to retrieve the header names from the JS config.

nemesisdesign (changes requested):
Instead of using "Loading new sessions..." can you use a loading indicator like the one used in notification widget?  Maybe the one we use can be reused, made smaller and positioned there, besides the "loading new sessions", on the left.

I tried with active sessions, the same width problem exists:

Screenshot from 2020-12-17 23-57-36

**Moreover, the bandwidth calculation is wrong, I just connected and it shows very high download and upload amounts, it's not possible, please calculate again.**

I tested with an online system, the requests take some instants to complete and the content in the page changes size after each request completes, which is not ideal, I think we should keep the loading overlay until the sessions are loaded and then hide it (`setLoading(false)`).

nemesisdesign (changes requested):
@NoumbissiValere great improvement the spinner! Some remaining issues with the last version, see this gif:

wifi-login-pages

### 1. Spinner appearing twice

The spinner appears twice intermittently.
This is confusing because it looks like the page has loaded but then the UI is blocked again.
Since we now have the nice spinner on the bottom while the sessions are loaded, can you reuse that spinner instead of showing the main spinner which blocks the UI?

So after the first spinner disappears, the spinner for the loading of sessions appears while sessions are being fetched.

### 2. The length of the sessions container is not right

The width of the session box is inconsistent with the main content, please make it consistent.

client/index.css
```diff
@@ -199,3 +228,20 @@ h1 {
...
+  animation: spin 0.8s linear infinite;
+}
+@keyframes spin {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
nemesisdesign, line 247: missing blank line

client/index.css
```diff
@@ -153,6 +154,34 @@ body .input.error {
...
 }
+td, th { 
+  padding: 1.9rem;
+}
+table {
+  background: #d3d3d3;
+  max-width: max-content;
+  margin-bottom: 20px;
+}
+.session-column {
```
nemesisdesign, line 165: please ensure that any CSS that is specific to the status component is in the CSS of the component instead of the general CSS, this is required because we'll implement lazy loading to load the minimum JS and CSS required by each page (that is, once a page is requested, its code is downloaded, while right now we load everything in one call).

nemesisdesign (changes requested):

nemesisdesign (changes requested):
Great work @NoumbissiValere! It's a lot better now! :+1:

I added two commits:

- 1st improves readability and white space, check it out
- 2nd makes the indentation in the CSS consistent with the rest of the codebase, which uses 2 spaces

There's only 1 remaining issue.

I tried logging in with a user which does not have any session.
**Expected outcome**: loading spinner disappears.
**Actual outcome**: loading spinner keeps spinning.

Please check and let me know, with that we can finally merge.

Thank you.

client/components/status/status.js
```diff
@@ -276,6 +279,15 @@ export default class Status extends React.Component {
     this.setState({screenWidth: window.innerWidth});
   }
 
+  updateSpinner = () => {
+    const {activeSessions, passedSessions} = this.state;
+    if (activeSessions.length === 0 && passedSessions.length === 0) {
+      this.setState({loadSpinner: false});
+    } else {
+      this.setState({loadSpinner: true});
+    }
```
nemesisdesign, line 288: what about:

```javascript
this.setState({
  loadSpinner: activeSessions.length || passedSessions.length
});
```

Or anything along these lines? The if/else is not really necessary since we're setting a boolean, the result of the set can be the if/else condition.

client/components/status/status.js
```diff
@@ -276,6 +279,15 @@ export default class Status extends React.Component {
     this.setState({screenWidth: window.innerWidth});
   }
 
+  updateSpinner = () => {
+    const {activeSessions, passedSessions} = this.state;
```
nemesisdesign, line 283: I didn't notice this before, `passedSessions` doesn't sounds like a good name to me, what about `pastSessions`? Please change all the occurrences.

client/components/status/status.test.js
```diff
@@ -592,4 +592,17 @@ describe("<Status /> interactions", () => {
...
+    prop.statusPage.links = links;
+    prop.isAuthenticated = true;
+    wrapper = shallow(<Status {...prop} />, {
+      context: {setLoading: jest.fn()},
+      disableLifecycleMethods: true,
+    });
+    // spinner should not load if no sessions are available
+    wrapper.instance().updateSpinner();
+    expect(wrapper.find(".loading").length).toEqual(0);
+
```
nemesisdesign, line 606: remove this blank line please

nemesisdesign (changes requested):
Good @NoumbissiValere, thanks for fixing that. I spotted a few minor things to improve which I commented below.

nemesisdesign (approved):

PR #152, opened 2021-01-15
[feature] Added more user fields #151

### LINKED ISSUE #151: [feature] Add more user fields in sign up page
Depends on https://github.com/openwisp/openwisp-radius/issues/219.

Allow the new fields (disabled by default) in the registration page, by configuration.

Each field can be: 

1. disabled (default)
2. allowed but optional (ensure the UI communicates the fact that the field is optional, we could use the HTML5 placeholder attribute)
3. mandatory

Ensure automated tests are present and a few lines in the README explain how to turn on each field.

NoumbissiValere:
Closes #151

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,15 +270,20 @@ With the configuration above:
 - the link to `twitter` (from Contact) and `change password` (from Header)
   links will be visible to only authenticated users
 
-#### Optional Fields
+#### User Fields in Registration Form
 
-We can use the `setting` attribute of optional fields `first_name`, `last_name`, `location` and `birth_date` 
-to decide if the fields should either be disabled, allowed but not required or required. `setting` can take 
-any of the following value:
+The `setting` attribute of the fields `first_name`, `last_name`,
+`location` and `birth_date` can be used to indicate whether the fields
+shall be disabled (the default setting), allowed but not required or required.
 
-- `disabled`: This is the default value. fields with this setting are not shown.
-- `allowed`: Fields with this setting are shown but not required.
-- `mandatory`: Fields with this setting are shown and required.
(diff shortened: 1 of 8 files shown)


nemesisdesign (commented):
I have one concern about the date field, is the format configurable?

I need to be able to set it to be: `day/month/year`, now is `month/day/year`, but this format depends on the language used so it's the only tricky part.

NoumbissiValere:
@nemesisdesign it is not possible to change the format of the `input` with `type=date`. I didn't see a problem with this because the value that is received by the `input` field irrespective of what is presented is `YYYY-MM-DD` according to [HTML5 date input specification](https://www.w3.org/TR/2011/WD-html-markup-20110405/input.date.html) which is the same format expected by the backend. I think the only way we can achieve this is by using an external tool but this will be extra memory just for date.  

With this, What do you suggest we do

nemesisdesign (commented):
> @nemesisdesign it is not possible to change the format of the `input` with `type=date`. I didn't see a problem with this because the value that is received by the `input` field irrespective of what is presented is `YYYY-MM-DD` according to [HTML5 date input specification](https://www.w3.org/TR/2011/WD-html-markup-20110405/input.date.html) which is the same format expected by the backend. I think the only way we can achieve this is by using an external tool but this will be extra memory just for date.
> 
> With this, What do you suggest we do

I double checked again and found out the format depends on the preferred language of the browser, I switched my browser preferred language from English to Italian and the format changes to `dd / mm / yyyy`, so that's cool, no action needed!

I will test again this asap and let you know if I find any issue. Thanks :+1:

nemesisdesign (approved):
:+1:

PR #166, opened 2021-02-03
added a loading animation

### LINKED ISSUE #156: [wifi-login-pages] Ensure something is shown while page loads
I've been testing the app in a real public wifi environment and when the page is loaded in the mini-browser of  android or ubuntu it takes too many seconds to show something, during the loading the page is completely blank.

We must make sure a basic layout with a text that explains the page is loading is shown to the user.
This basic layout can be built in the main HTML file, without relying on external files (we can use the `<style>` tag for CSS).

**Update**: I think we should simply show the loading indicator, embedding it in public/index.html with inline code.
We shouldn't use words because the pages can be displayed in multiple languages and we have no way of translating the word in HTML without loading JS, therefore we should resort to something visual only.

vivek-30:
Added a custom loading animation which fixes #156

### DIFF
--- a/client/index.css
+++ b/client/index.css
@@ -14,8 +14,9 @@
 }
 
 #preload{
-  display:none;
+  display: none;
 }
+html, body{ overflow: initial }
 
 body {
   margin: 0;
--- a/client/index.css
+++ b/client/index.css
@@ -12,6 +12,11 @@
   font-display: swap;
   src: local('Montserrat SemiBold'), local('Montserrat-SemiBold'), url(assets/fonts/montserrat-semi-bold.woff2) format('woff2');
 }
+
+#preload{
+  display:none;
+}
+
 body {
   margin: 0;
   padding: 0;
--- a/public/index.html
+++ b/public/index.html
@@ -6,14 +6,19 @@
     <link rel="shortcut icon" href="favicon.png">
     <title>OpenWISP WiFi Login</title>
   </head>
-
   <style>
+    html, body{ overflow: hidden }
     #preload {
-      width: 60px;
-      height: 60px;
-      position: absolute;
-      top:50%;
-      left:50%;
+      width: 100%;
+      height: 100%;
+      position: fixed;
+      top: 0;
+      left: 0;
+      background: rgba(255, 255, 255, 0.95);
+      z-index: 9999;
+      display: flex;
+      justify-content: center;
+      align-items: center;
     }
     .loader {
       border: 8px solid #f3f3f3;
@@ -29,7 +34,6 @@
       100% { transform: rotate(360deg); }
     }
   </style>
-
   <body>
     <div id="preload">
       <div class="loader"></div>
--- a/public/index.html
+++ b/public/index.html
@@ -6,7 +6,34 @@
     <link rel="shortcut icon" href="favicon.png">
     <title>OpenWISP WiFi Login</title>
   </head>
+
+  <style>
+    #preload {
+      width: 60px;
+      height: 60px;
+      position: absolute;
+      top:50%;
+      left:50%;
+    }
+    .loader {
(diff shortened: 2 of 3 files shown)


vivek-30:
Sir this is the `working gif` for mentioned problem -

loading_animation

vivek-30:
@nemesisdesign , sir could you please review this.

public/index.html
```diff
@@ -6,7 +6,51 @@
...
+    50% { transform: rotate(180deg); border-width: 1px; }
+    100% { transform: rotate(360deg); border-width: 6px; }
+    }
+
+    @keyframes circle-two {
+      0% { transform: rotate(0deg); border-width: 1px; }
+      50% { transform: rotate(180deg); border-width: 6px; }
+      100% { transform: rotate(360deg); border-width: 1px; }
+    }
+  </style>
```
nemesisdesign, line 47: we already have styles for the loader, is it possible to move these to the style tag so they can be reused across the whole application and avoid redefining the same styles multiple times?

See (except `.org-wrapper-not-found`):
https://github.com/openwisp/openwisp-wifi-login-pages/blob/50c2885200fe5b4369b119d0608bb03222eb5031/client/components/organization-wrapper/index.css#L17-L57

client/index.css
```diff
@@ -12,6 +12,11 @@
   font-display: swap;
   src: local('Montserrat SemiBold'), local('Montserrat-SemiBold'), url(assets/fonts/montserrat-semi-bold.woff2) format('woff2');
 }
+
+#loading{
```
nemesisdesign, line 16: can we rename this something like `#preload`?

nemesisdesign (changes requested):
Thank you @vivek-30!

It takes some time for us to review PRs, thank you for your patience.

I tested it and I can confirm this is what we need, however we can do some improvements to avoid code duplication, see my comment below.

client/index.css
vivek-30: Ok sir , i will change it to `preload`  as it will make  more sense ☺️

public/index.html
vivek-30: Sir, one things i would like to ask that do we need to shift some of style tag code to index.css for their reusability? Is this what you are asking for?

client/index.css
vivek-30: ok i got this . we can use `loader` class for `#preload div` and can then add some remaining style property likewise we can use `spin` animation of index.css file but i am wondering how we can use them without involving javascript?

May be we can render this preload div using react and can then include these styles but then the problem is still remain unsolved due to delay in render of these javascript file(s)

vivek-30: what should we do sir any suggestions ?

nemesisdesign: I think the HTML should be kept as is and the CSS should be moved from the JS file to index.html in the style tag. Try that, it should work, I tried it quickly before suggesting.

nemesisdesign: please also remember to run `./run-qa-checks`. The CI build is not working now but it will be resumed soon.

vivek-30: @nemesisdesign , Thank a lot sir for your help now i have done required changes and also ran `./run-qa-checks` which you have suggested me.
And also Everything is fine from my side could you please confirm it ☺️.

vivek-30:
@nemesisdesign sir could you please review these changes ☺️

nemesisdesign (approved):
Thanks @vivek-30 :+1:

vivek-30:
@nemesisdesign if there is any other issue which i can work on or you want help on please let me know.
I will be happy to do that 😊 .Thank a lot sir BTW in solving this problem.

nemesisdesign:
@vivek-30 thank you! There's plenty of things you can help us with, apart from the issues in this repository, you can also take a look at the project board: https://github.com/orgs/openwisp/projects/4

PR #159, opened 2021-01-30
[enhancement] Implemented lazy loading to speed up first load #154

niteshsinha17:
closes #154

PR #170, opened 2021-02-12
[feature] Added option for advanced signup

NoumbissiValere:


nemesisdesign (changes requested):
I get this error when I open the sign up of the default org:

```
TypeError: Cannot read property 'replace' of undefined
[0]     at some (/openwisp-wifi-login-pages/server/controllers/getPlans.js:16:33)
[0]     at Array.some (<anonymous>)
[0]     at getPlans (/openwisp-wifi-login-pages/server/controllers/getPlans.js:9:30)
[0]     at Layer.handle [as handle_request] (/openwisp-wifi-login-pages/node_modules/express/lib/router/layer.js:95:5)
[0]     at next (/openwisp-wifi-login-pages/node_modules/express/lib/router/route.js:137:13)
[0]     at Route.dispatch (/openwisp-wifi-login-pages/node_modules/express/lib/router/route.js:112:3)
[0]     at Layer.handle [as handle_request] (/openwisp-wifi-login-pages/node_modules/express/lib/router/layer.js:95:5)
[0]     at /openwisp-wifi-login-pages/node_modules/express/lib/router/index.js:281:22
[0]     at Function.process_params (/openwisp-wifi-login-pages/node_modules/express/lib/router/index.js:335:12)
[0]     at next (/openwisp-wifi-login-pages/node_modules/express/lib/router/index.js:275:10)
```

Can you include in the PR text basic instructions to test it? I can't see the new flow, probably the error above is due to a missing configuration, although it would be better to loudly indicate the error.

It's also not entirely clear the following:

- is the feature opt-in only? It must be disabled by default
- is there a success and failure page/screen?

NoumbissiValere:
> It's also not entirely clear the following:
> 
> * is the feature opt-in only? It must be disabled by default
> * is there a success and failure page/screen?

It is disabled when there no plans available. You can go ahead and test it again. to test the functionality of plans, you need to make it work with this branch https://github.com/openwisp/openwisp-subscriptions/tree/issue-5-add-gestpay-service of openwisp-subscriptions

nemesisdesign (changes requested):
Great progress Noumbissi!

See my findings below.
 
## Sign up page plan loading

When I open the sign up page, I briefly see the fields while plans load.
This should not happen, the loading indicator (with `setLoading` or anything similar) should be shown while plans are loaded.

## Status page being shown after sign up

Once I sign up after selecting a plan, I get logged in and the status page is shown while I'm being redirected to gestpay, this should not happen.

## What about SMS verification?

If you remember our past discussions, I explained that right now the most urgent case is the following:

- users either register with their national phone number
- users who do not have a national phone number will use the credit card

This at the moment is not possible with the current implementation, how do you intend to implement it?

## Redundant fields

There are some redundant fields, for example, full name and first_name/last_name.

We should make first_name and last_name mandatory in openwisp-subscriptions when activating the subscriptions module for an org, full name can be filled from that field.

We can also automatically fill User.location with the billing address.

## Latest master

Please also rebase on the latest master, ensure lazy loading is respected (use it when it allows to avoid including files in the first webpack build, see the changes that have been introduced in https://github.com/openwisp/openwisp-wifi-login-pages/pull/159).

## Basic docs

If you could add a basic section with brief instructions about how to enable this feature it would be great, I will help you to add more info and make it more complete.

NoumbissiValere:
> ## What about SMS verification?
> If you remember our past discussions, I explained that right now the most urgent case is the following:
> 
> * users either register with their national phone number
> * users who do not have a national phone number will use the credit card
> 

@nemesisdesign I think with the `verifies_identity` feature available this is no problem because credit card verification have precedence over sms verification. 

* If the plan selected has verifies_identity true, credit card will be verified during signup and not phone_number. attention is given to credit card and not phone number
* If the plan selected has verifies_identity false, signup flow works ask before.

So if a user does not have a national phone_number:
* and the selected plan has verifies_identity true, credit card of that user is verified.
* Or the selected plan has verifies_identity false, signup falls back to what the system had before. we can't fall back to credit card verification because, verifies_identity for the selected plan is false.

server/controllers/getPlans.js
```diff
@@ -0,0 +1,66 @@
+import axios from "axios";
+import merge from "deepmerge";
+import config from "../config.json";
+import defaultConfig from "../utils/default-config";
+import logInternalError from "../utils/log-internal-error";
+
+const getPlans = (req, res) => {
+    const reqOrg = req.params.organization;
+    const validSlug = config.some(org => {
+        if (org.slug === reqOrg) {
```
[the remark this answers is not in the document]
NoumbissiValere: @nemesisdesign this server script is only called when subscription module is enabled.

nemesisdesign (changes requested):
Good progress, I see that when the identity verification flag is not checked the fields for the invoice are not shown and if the SMS verification is enabled the flow falls back to that which is good.

See the problems I found during testing below.

## Country restriction

As explained in our discussions, the use case we're implementing now works as follows:

- users with a national phone number will register using their phone number (only national prefixes are allowed by the phone input)
- all other users will use credit card verification without phone number,
  since they do not have a national phone number, they need to be able to log in with the username

With the current implementation, the phone number is asked for anyway in the sign up,
but since the configuration will be limited to a specific prefix,
users which do not have a national SIM Card will not be able to register or may enter a fake phone to register.

The other problem is that when logging in, the phone number is asked. But the phone input must be limited to the national prefixes, so users can't enter a phone number.

Did you think about this when designing the solution and estimating the issue?
What solution you had in mind?

Let's discuss a bit more before jumping to the implementation.

## Success and failure URLs

What are the payment success and payment failure URLs?

Once a payment succeeds or fails, openwisp-subscriptions will redirect to specific URLs.

Now the URLs redirect to:

- http://localhost/captivepage/success.html
- http://example.com/success/
- http://localhost/captivepage/failure.html

Can you please indicate in the README of openwisp-wifi-login-pages what are the URLs
to use in openwisp-subscriptions?

## More than 3 plans are not shown

I tried adding a new plan to openwisp-subscriptions (apart from the default ones)
but it was not shown until I deleted one of the defaults, do you know why this is happening?

nemesisdesign (changes requested):
## Status page

Initialize the signup process, create the user, interrupt the payment, open the status page again.

The status page shows "WiFi Login Successful!", but the sign up process is not completed yet.

We need a way to know if the user has completed the payment or not and if not we should show a different message.

client/components/registration/registration.js
```diff
@@ -54,17 +95,25 @@ export default class Registration extends React.Component {
   handleSubmit(event) {
     const {setLoading} = this.context;
     event.preventDefault();
-    const {orgSlug, authenticate, settings, setUserData, userData} = this.props;
+    const {orgSlug, authenticate, settings} = this.props;
```
codesankalp, line 98: I think this thing breaks the signup process for the organization without any verification. I have tested it for default organization.

client/components/status/status.js
```diff
@@ -87,65 +90,84 @@ export default class Status extends React.Component {
...
+      }
+
+      if (needsVerify("bank_card", userData, settings)) {
+        window.location.assign(userData.payment_url);
+      }
+
+      if (userData.is_verified) {
+        setLoading(false);
+      } else {
+        return;
```
codesankalp, line 160: Instead of return why don't we logout the user?

client/components/status/status.js
```diff
@@ -87,65 +90,84 @@ export default class Status extends React.Component {
...
+        this.loginFormRef.current.submit();
+        userData.justAuthenticated = false;
+        setUserData(userData);
+      }
+
+      if (needsVerify("bank_card", userData, settings)) {
+        window.location.assign(userData.payment_url);
+      }
+
+      if (userData.is_verified) {
```
codesankalp, line 157: If the organization doesn't support verification thensetLoading is always true. I tested it with default org and the loader is on the screen.

client/components/status/status.js
nemesisdesign: because the user needs to do the verification process and needs to be logged in to do that

client/components/status/status.js
nemesisdesign: oh right. So it should check whether any type of verification is enabled.

client/components/registration/registration.js
nemesisdesign: ok thank you I think I did not test that, I will double check and add a failing test

client/components/status/status.js
codesankalp: Yes

client/components/status/status.js
codesankalp: Yeah, I understood.

nemesisdesign (commented):
@codesankalp https://github.com/openwisp/openwisp-wifi-login-pages/pull/170/commits/b854db5b3cf747b621cd9b180de7b65735dda706 fixes email registration, I did some manual testing and I added a failing test in `status.tests.js`, I took advantage to add a few basic tests for simple registration in `registration.tests.js` although those do not fail, I added those just for completing the test suite.

nemesisdesign (commented):
QA checks passed, let's wait no more and merge so you can rebase your open PRs and we can move on, I'll do further improvements in the next days.

Issue #153, opened 2021-01-22
[CI] Move CI to github actions


nemesisdesign:
Move CI to github actions.

sankha555:
I had read upon Github actions a few weeks back. I'll try to make a patch for this.

devkapilbansal:
@sankha555 are you working on this?

sankha555:
I am not able to contribute currently, it can please be assigned to someone else. Sorry for the inconvenience.

codesankalp:
Can I work on this issue? @nemesisdesign @devkapilbansal

nemesisdesign:
Whoever wants to work on this, please go ahead and create a PR without asking for permission.

PR #174, opened 2021-02-23
[change] Moved CI to GitHub actions #153

nemesisdesign:
CC: @codesankalp


.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
+---
+name: OpenWisp WiFi Login Pages CI BUILD
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches: -master
```
nemesisdesign, line 9: is the dash before `master` the reason for which the build doesn't start?

PR #181, opened 2021-03-10
[footer] Fix unique key props error #169

### LINKED ISSUE #169: [bug] Fix unique key props error when no links
Whenever there is no link present in the footer unique key props error is shown in the browser console.

s0mnaths:
Just had to change the key. 
Earlier the key was the URL itself, and when no link was present in the footer, a unique key prop error would come.

Fixes #169

### DIFF
--- a/client/components/footer/footer.js
+++ b/client/components/footer/footer.js
@@ -24,7 +24,7 @@ export default class Footer extends React.Component {
                   footer-link-${index + 1}`}
                     target="_blank"
                     rel="noreferrer noopener"
-                    key={link}
+                    key={index}
                   >
                     {getText(link.text, language)}
                   </a>
--- a/client/components/footer/footer.js
+++ b/client/components/footer/footer.js
@@ -24,7 +24,7 @@ export default class Footer extends React.Component {
                   footer-link-${index + 1}`}
                     target="_blank"
                     rel="noreferrer noopener"
-                    key={link.url}
+                    key={link}
                   >
                     {getText(link.text, language)}
                   </a>


nemesisdesign (commented):
We need at least 1 test.

@niteshsinha17 what was the issue? Do you remember?

client/components/footer/footer.js
```diff
@@ -24,7 +24,7 @@ export default class Footer extends React.Component {
                   footer-link-${index + 1}`}
                     target="_blank"
                     rel="noreferrer noopener"
-                    key={link.url}
+                    key={link}
```
niteshsinha17, line 27: Can we set ``key=index``? 
As link is an object we will again get unique key props error. 
What do you think?

niteshsinha17:
@nemesisdesign 
Actually error is caused when we have no links to display in the footer. But we get an empty object inside ``links`` which don't have any ``url``. If we can somehow check console error in the test then we can add test for it.

niteshsinha17:
Also, we are going to write test for it then we need to test two cases.
1. When there is no link.
2. When there are more than two links with ``authenticated=true``.

nemesisdesign:
Thanks @niteshsinha17! @s0mnaths let us know if it's clear.

s0mnaths:
> Thanks @niteshsinha17! @s0mnaths let us know if it's clear.

Yup it's clear now. I'm just figuring out how to write the test. Will try to update you guys asap. Thank you!

frappelatte28:
https://github.com/openwisp/openwisp-wifi-login-pages/blob/aa7f135b75440fd9a6ce1c1797d5d11ff0ad97d0/client/components/footer/footer.test.js#L53

I am not sure but dont we already have tests written for these cases @niteshsinha17 ?

niteshsinha17:
> https://github.com/openwisp/openwisp-wifi-login-pages/blob/aa7f135b75440fd9a6ce1c1797d5d11ff0ad97d0/client/components/footer/footer.test.js#L53
> 
> I am not sure but dont we already have tests written for these cases @niteshsinha17 ?

No, this test is not for these cases. It is just testing that component should render when there are no links. I would like to mention the current case again. When we are having no links then still we have an empty object in place of the link like this ``[{}]``. I hope this will make you guys to understand the situation. Also, I would suggest you to find out why we are getting this empty object here. This will help you to understand the code.

s0mnaths:
@niteshsinha17 @nemesisdesign  Can you please help me with this commit? Why are the checks failing and how can they be fixed?

Also, I could only do one of the changes requested. I got familiar with writing tests using react-test-render and enzyme but couldn't figure out how to make sure the key props remain unique. Since this is a very small issue, can't we accept the patch without a test?

codesankalp:
image

The QA-Checks are failing due to the es-lint -> `array index`  use in the footer.

Also, @s0mnaths writing test is good even for a small change because it will increase code coverage which is the most important thing here.

s0mnaths:
> image
> 
> The QA-Checks are failing due to the es-lint -> `array index` use in the footer.
> 
> Also, @s0mnaths writing test is good even for a small change because it will increase code coverage which is the most important thing here.

Oh, I got it. Thank you! And yes I understand the need for writing tests here but I still can't figure out how to test for unique props. 

@niteshsinha17 since we shouldn't use `key={index}` here, should we use `key={links}` or try to find some other way?

niteshsinha17:
> Also, I could only do one of the changes requested. I got familiar with writing tests using react-test-render and enzyme but couldn't figure out how to make sure the key props remain unique. Since this is a very small issue, can't we accept the patch without a test?

Don't worry I will dedicate some time on this Issue to help you.

niteshsinha17:
It looks like this error is coming from here.
https://github.com/openwisp/openwisp-wifi-login-pages/blob/b71dceaf5846763b18fb54b24980e4ff4475687e/internals/generators/organization/config.yml.hbs#L88-L91 
We can solve this Issue by changing it like this-
```
footer: 
       links: []
```
@nemesisdesign Can you please confirm about this change. Was there any intention of writing it like that.

niteshsinha17:
@nemesisdesign I have tested it that I mentioned in my last comment. It's working fine.

s0mnaths:
> @nemesisdesign I have tested it that I mentioned in my last comment. It's working fine. I think we can change it

Okay then I will do it, @nemesisdesign is that okay?

nemesisdesign (commented):
What's the status of this PR?

Can we please wrap up and finish it?

@s0mnaths follow the guidance of @niteshsinha17.

nemesisdesign:
It's going to be handled in #201. Thanks.

PR #184, opened 2021-03-16
[fix] Fixes unique key props error

frappelatte28:
Problem was key was being set to inner link it should be on outer div so when multiple renders occur react will know the key is same or not or it has to render element or not.

@niteshsinha17 Please review this PR.

### DIFF
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -537,10 +537,9 @@ export default class Status extends React.Component {
               {links && links.map(link => {
                 if (shouldLinkBeShown(link, isAuthenticated)) {
                   return (
-                    <div className="links row">
+                    <div className="links row" key={link}>
                       <Link
                         className="button full status-link"
-                        key={link.url}
                         to={link.url.replace("{orgSlug}", orgSlug)}
                       >
                         {getText(link.text, language)}
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -537,7 +537,7 @@ export default class Status extends React.Component {
               {links && links.map(link => {
                 if (shouldLinkBeShown(link, isAuthenticated)) {
                   return (
-                    <div className="links row" key={link}>
+                    <div className="links row" key={link.url}>
                       <Link
                         className="button full status-link"
                         to={link.url.replace("{orgSlug}", orgSlug)}


niteshsinha17:
Thanks @devkapilbansal  and @frappelatte28. I will review your PR today.

niteshsinha17:
@frappelatte28  You should follow the instructions mentioned here before opening a PR. https://openwisp.io/docs/developer/contributing.html

frappelatte28:
Yes, i'll keep it in mind @niteshsinha17 . I read these guidelines a bit later after creating a PR then I made the changes have I made any mistakes in the last commit too? Just let me know my mistakes and i'll make sure it won't happen again.

niteshsinha17:
I think you have done mistake during creating PR. You should check title and comment box carefully.

client/components/status/status.js
```diff
@@ -537,10 +537,9 @@ export default class Status extends React.Component {
               {links && links.map(link => {
                 if (shouldLinkBeShown(link, isAuthenticated)) {
                   return (
-                    <div className="links row">
+                    <div className="links row" key={link}>
```
niteshsinha17, line 540: This is not perfect solution. You will get ``` Encountered two children with the same key, `[object Object]``` warning for multiple objects.

niteshsinha17 (changes requested):

client/components/status/status.js
niteshsinha17: Take reference from this conversation. https://github.com/openwisp/openwisp-wifi-login-pages/pull/181#discussion_r592025358

frappelatte28: @niteshsinha17  
It would nice to hear some suggestions on facts,  on grounds of which i conducted this approach :-
* For multiple render() , key for same element must be same ( just to not create that element multiple times ) ?
* Also  key must present on outer element otherwise the react will take outer element as new element every time unless you provide a key to it to make react understand that okay this is same last element and react wont recreate it in dom ?

also
I tried this approach and i did not get any such error , am i missing something to reproduce the error you mentioned ?

niteshsinha17: 1. We can not have the same key for different elements wheater they are the same or different. 

2. Yes, key must be present on the outer element. This was the only reason for which we were getting key warning. 

3.In this situation, you can use index or url as the key. 
I also understand we should not use the index as the key but this is true when the list gets changed every time. Here we are showing a button on the status component with the help of a list that may not change again and again. 

4. Also you can't reproduce that warning here. But see the image below. It will help you to understand.

Screenshot (164)

frappelatte28: @niteshsinha17  thank you for the explanation.
for the last point , i think this is the case when we would have same `links` which i dont think we would? , so for different `link` we would have different key .

Also, if i am able to  sum up correctly ,  only thing left is to assign `key={link.url}` at  outer div ?

niteshsinha17: Yes, you are right

frappelatte28:
I have made the changes @niteshsinha17 :-)

nemesisdesign (approved):
Since it's a small issue I can accept the patch without a test, but keep in mind that writing tests is a key requirement for contributors and GSoC participants, so you will need to demonstrate you're able to master this skill.

nemesisdesign:
@niteshsinha17 does the fix work as expected?

codesankalp (approved):
image

Tested, Working fine.

niteshsinha17:
> @niteshsinha17 does the fix work as expected?

Yes

niteshsinha17 (approved):

frappelatte28:
> Since it's a small issue I can accept the patch without a test, but keep in mind that writing tests is a key requirement for contributors and GSoC participants, so you will need to demonstrate you're able to master this skill.

Yes sure @nemesisdesign. I have started learning the protocol and will keep that in my mind going forward.

Issue #171, opened 2021-02-13
[wifi-login-pages] Logout without breaking auto-login


nemesisdesign:
To be done after #63.

When a user logs out, once they go back to the log in page, they will not be auto logged in if they enabled the remember me checkbox.

We need to find a way to make the auto-login feature work when the remember-me checkbox is enabled, so that when they connect to the WiFi again, they'll be automatically logged in when they open the page.

Right now users get automatically logged in only if they did not click the logout button.

The problem is that right now if one clicks logout, they're taken back to the login page, and without clearing the cookie session, they would be immediately logged in again.

So we could implement something like the following:

- if the remember me checkbox is enabled and the user logs out, ask them (with a modal message) if they want to be automatically logged in the next time they open the wifi
- if they choose no, the behavior is unchanged
- if they choose yes, the cookie/session data is not erased, but the captive portal log out is performed normally
- once the log out is performed, instead of taking back the user to the log in page, we direct them at a new page or a screen in the same status page, which communicates the successful log out and shows a button that says "Log in again" or something similar, which links the log in page

jalajcodes:
I'd like to work on this issue.

codesankalp:
I can work on this.

codesankalp:
@nemesisdesign Since I have made a PR for [this](https://github.com/openwisp/openwisp-wifi-login-pages/issues/63).
Can I start working on this issue?

nemesisdesign:
@codesankalp yes sure, is the problem clear?

codesankalp:
> @codesankalp yes sure, is the problem clear?

Yes, I think so. If I will face any problem related to implementation I will ask it here.

PR #186, opened 2021-03-18
[wifi-login-pages] Added remember me checkbox #63

### LINKED ISSUE #63: [wifi-login-pages] Add remember me checkbox
Add a "remember me" checkbox, enabled by deafult.

If the checkbox is disabled, the app will not store session information in cookies (as it does now by default), but it will use `sessionStorage` as an alternative way to recognize that users have already logged-in (showing them the status page).

This feature is important because public wifi is sometimes used in shared computers in public places, and therefore we may have to disable it by default.

codesankalp:
Added remember_me checkbox with sessionStorage.

Closes #63

### DIFF
--- a/client/components/login/login.js
+++ b/client/components/login/login.js
@@ -35,6 +35,7 @@ export default class Login extends React.Component {
       email: "",
       password: "",
       phone_number: "",
+      remember_me: true,
       errors: {},
     };
     this.handleSubmit = this.handleSubmit.bind(this);
@@ -44,6 +45,14 @@ export default class Login extends React.Component {
   componentDidMount() {
     const email = getParameterByName("email");
     const token = getParameterByName("token");
+
+    const {loginForm} = this.props;
+    let remember_me;
+    if (localStorage.getItem("rememberMe") !== null)
+      remember_me = localStorage.getItem("rememberMe") === "true";
+    else remember_me = loginForm.input_fields.remember_me.value;
+    this.setState({remember_me});
+
     if (email && token) {
       this.setState(
         {
@@ -147,15 +156,21 @@ export default class Login extends React.Component {
     const {setLoading} = this.context;
     if (event) event.preventDefault();
     const {orgSlug, authenticate, verifyMobileNumber, settings} = this.props;
-    const {email, password, phone_number, errors} = this.state;
+    const {email, password, phone_number, remember_me, errors} = this.state;
     const username = settings.mobile_phone_verification ? phone_number : email;
     const url = loginApiUrl(orgSlug);
     this.setState({
       errors: {},
     });
+    localStorage.setItem("rememberMe", remember_me);
     setLoading(true);
 
-    const handleAuthentication = function (needsMobileVerification = false) {
+    const handleAuthentication = (
+      needsMobileVerification = false,
+      data = {},
+    ) => {
+      if (!remember_me)
+        sessionStorage.setItem(`${orgSlug}_auth_token`, data.key);
       authenticate(true);
       toast.success(loginSuccess, {
         toastId: mainToastId,
@@ -177,8 +192,11 @@ export default class Login extends React.Component {
         password,
       }),
     })
-      .then(() => {
-        return handleAuthentication();
+      .then((res = {}) => {
+        return handleAuthentication(
+          settings.mobile_phone_verification,
+          res.data,
+        );
       })
       .catch((error) => {
         const {data} = error.response;
@@ -205,8 +223,14 @@ export default class Login extends React.Component {
       });
   }
 
+  handleCheckBoxChange = (event) => {
+    this.setState({
+      remember_me: event.target.checked,
+    });
+  };
+
   render() {
-    const {errors, password} = this.state;
+    const {errors, password, remember_me} = this.state;
     const {
       language,
       loginForm,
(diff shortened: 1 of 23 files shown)


codesankalp:
@nemesisdesign, @niteshsinha17 Can you review it?
I tried to implement the sessionStorage when `remember_me` is unchecked and made it compatible with all other previous code.

niteshsinha17:
@codesankalp You have formatted lots of files. It made it difficult to review also to understand which change is made for this particular issue.

codesankalp:
@niteshsinha17 I will unstage the reformatted file in which I haven't added a code for this implementation.

codesankalp:
@niteshsinha17 I have removed the extra reformatted files.

Also, I have implemented the remember_me feature only for the login.
If it's required for registration then do let me know.

codesankalp:
@nemesisdesign Can you please review it?
I am using commit of this PR in https://github.com/openwisp/openwisp-wifi-login-pages/issues/171.
And will submit a PR soon.

nemesisdesign:
@codesankalp sorry, I need more days.

client/components/login/login.js
```diff
@@ -59,8 +60,8 @@ export default class Login extends React.Component {
   }
 
   getEmailField = (input_fields) => {
-    const {email, errors} = this.state;
-    const {language} = this.props;
+    const { email, errors } = this.state;
+    const { language } = this.props;
```
nemesisdesign, line 64: please avoid these kind of style changes. If you want to add an automatic formatting tool that does this kind of change, please open a new issue and a new PR just for that.

This applies to any change in this PR which is not directly related to the feature you're working on.

internals/generators/organization/config.yml.hbs
```diff
@@ -422,6 +422,10 @@ client:
             en: "password"
           label:
             en: "password"
+        remember_me:
+          type: "checkbox"
+          label:
+            en: "Remember Me"
```
nemesisdesign, line 428: let's change this to "Automatically log me in to the WiFi service at the next visit" which I hope most end-users will be able to understand

org-configurations/default-configuration.yml
```diff
@@ -159,7 +159,7 @@ client:
           pattern_description:
             en: "only letters, numbers, and @/./+/-/_ characters"
           placeholder:
-            en: "enter location" 
+            en: "enter location"
```
nemesisdesign, line 162: avoid this unrelated change please

client/components/login/login.test.js
```diff
@@ -110,6 +110,15 @@ describe("<Login /> interactions", () => {
...
 
+  it("should change state value when handleCheckBoxChange function is invoked", () => {
+    // phone_number should not be present if mobile_phone_verification is off
+    expect(wrapper.find("[name='phone_number']")).toEqual({});
+    wrapper
+      .find("#remember_me")
+      .simulate("change", { target: { checked: false, name: "remember_me" } });
+    expect(wrapper.state("remember_me")).toEqual(false);
+  });
+
```
nemesisdesign, line 121: Good that you added one test, although looks not sufficient to prevent this feature from breaking.
Is there any measure you can include in this PR so that if someone changes the code and breaks this feature, the tests will find out?

nemesisdesign (changes requested):
Sorry for my delay @codesankalp, here's my review.

First of all, thank you for your effort, I tested it and it works! 
We need to improve it a bit, see below.

### Remember me choice should be maintained

If I choose to opt out from "remember me" and I close and then open the browser again, the checkbox should remain unchecked because otherwise the user could inadvertently log-in with the option activated from a shared computer. Can you please do this change (we can use localStorage)?

### Configurable default value

Can you make the default value of the checkbox configurable?
Some orgs may want to have this turned on by default while others off by default.

codesankalp:
Sorry for the delay @nemesisdesign Most of the work is done but needs some more time in adding tests for session storage so that this feature can be tested well and it will not break again in the future.

codesankalp:

internals/generators/organization/index.js
```diff
@@ -85,6 +85,12 @@ module.exports = {
       message: "Does your captive portal support log out by session ID?",
       default: false,
     },
+    {
+      type: "confirm",
+      name: "remember_me",
+      message: "Does your captive portal support automatically log me in?",
```
nemesisdesign, line 91: mm.. did I ask you to add this?

If I am not missing anything, this is not a feature of the captive portal but something we are adding on top, because we save the session of the user from which we're then able to retrieve the radius user token and log in the user automatically.

So the question should be something like: `Do you want to support automatic logins (remember me checkbox)?`

org-configurations/default-configuration.yml
```diff
@@ -340,6 +340,11 @@ client:
             en: "password"
           label:
             en: "password"
+        remember_me:
+          type: "checkbox"
+          value: true
+          label:
+            en: "Automatically log me in to the WiFi service at the next visit"
```
nemesisdesign, line 347: after testing it I noticed it's too long, can you please shorten this and the other occurrences to something more conservative like: `Remember me on this device`?

nemesisdesign (changes requested):
Thank you @codesankalp, this is almost ready, see below for my final requests for improvement.

codesankalp:
@nemesisdesign Should I squash the commits?

nemesisdesign:
> @nemesisdesign Should I squash the commits?

No, don't worry, I'll do it.

nemesisdesign (approved):
Thanks! :+1:

PR #199, opened 2021-04-15
[QA] Fixed coveralls to upload coverage #188

### LINKED ISSUE #188: [QA] Fix Coveralls to upload coverage
CI build of PR shows that coveralls is not working fine.

image

codesankalp:
Added yarn coverage and removed yarn test to generate coverage file.

Fixes #188


codesankalp:
a
`ref: ${{ github.event.pull_request.head.sha }}`
Due to this I have to make separate files @nemesisdesign

nemesisdesign:
I do not understand very well why is this needed, just for the record could you please explain a bit more in detail @codesankalp?

codesankalp:
@nemesisdesign In previous builds coveralls can't find the coverage file because it's not present.
image

When I used `yarn coverage` i.e. `yarn test --coverage` now it makes the coverage file.
But there is still a problem in uploading coverage to coveralls.

Now as the sha of the merge commit is used as reference here the coveralls app which uses `git cat-file -p <sha>` shows invalid name because `sha` is provided instead of `commit hash` as shown in the image.
a

As head_ref (`${{ github.event.pull_request.head.sha }}`) is required to check commit message. I have made a separate file yesterday.

I tried with a different `ref` but it didn't work for commit message check.
So, I decided to make them separate.

devkapilbansal:
@codesankalp what's the need of breaking workflow in two files

codesankalp:
> @codesankalp what's the need of breaking workflow in two files

I think I have explained it in the above comment.
Simply, If the ref is from `${{ github.event.pull_request.head.sha }}` then coveralls shows invalid object name due to this command `git cat-file -p <id>` but if I remove this `${{ github.event.pull_request.head.sha }}` then commit message check fails. So I have to split it into two files one is for QA-checks and the other is for test and coverage.

PR #201, opened 2021-04-19
[bug] Fixed unique key props error when no links #169

### LINKED ISSUE #169: [bug] Fix unique key props error when no links

codesankalp:
Added check for undefined link url

Closes #169

### DIFF
--- a/client/components/footer/footer.js
+++ b/client/components/footer/footer.js
@@ -17,30 +17,31 @@ export default class Footer extends React.Component {
       <div className="footer-container">
         <div className="footer-row-1">
           <div className="footer-row-1-inner">
-            {links.map((link) => {
-              if (shouldLinkBeShown(link, isAuthenticated, userData)) {
-                const css = link.css || "";
-                return (
-                  <a
-                    href={link.url}
-                    className={`footer-link ${css}`}
-                    target="_blank"
-                    rel="noopener noreferrer"
-                    key={link.url}
-                  >
-                    {link.icon && (
-                      <img
-                        src={getAssetPath(orgSlug, link.icon)}
-                        alt={link.alt ? getText(link.alt, language) : ""}
-                        className="footer-image"
-                      />
-                    )}
-                    {link.text ? getText(link.text, language) : ""}
-                  </a>
-                );
-              }
-              return null;
-            })}
+            {links &&
+              links.map((link) => {
+                if (shouldLinkBeShown(link, isAuthenticated, userData)) {
+                  const css = link.css || "";
+                  return (
+                    <a
+                      href={link.url}
+                      className={`footer-link ${css}`}
+                      target="_blank"
+                      rel="noopener noreferrer"
+                      key={link.url}
+                    >
+                      {link.icon && (
+                        <img
+                          src={getAssetPath(orgSlug, link.icon)}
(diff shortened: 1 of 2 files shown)


codesankalp:

nemesisdesign (commented):
I have one doubt: how to replicate the issue in the first place?

I tried setting:

```
    footer:
      links: []
```

In the config but I don't get any error. I don't understand why this change is necessary.
There's no point having a link which doesn't have `url` set, or does it and if yes why? I lost track of this issue.

codesankalp:
@nemesisdesign in the config generator i.e.
https://github.com/openwisp/openwisp-wifi-login-pages/blob/3934656624fd465939433dfd22b242529775525f/internals/generators/organization/config.yml.hbs#L88-L90
when creating a new org new config file is being generated with empty links i.e. the URL is not present in the generated config file which gives a unique key props error as URL is not present.
So, Instead of using `index` as a key, I checked whether a URL is present in the link, if it's present it will be a unique key and if it's not present then simply it will return null. And it works for other cases also (because without a URL there is no need for `<a>` tag.).

In the comment of @niteshsinha17 [here](https://github.com/openwisp/openwisp-wifi-login-pages/pull/181#issuecomment-805086283), he suggested using `links: [ ]` but if I will use this then it will be hard to get the text for the link.

niteshsinha17:
>he suggested using links: [ ] but if I will use this then it will be hard to get the text for the link.

@codesankalp Can you please explain why do we need a text without any link in the footer links section?

codesankalp:
> > he suggested using links: [ ] but if I will use this then it will be hard to get the text for the link.
> 
> @codesankalp Can you please explain why do we need a text without any link in the footer links section?

@niteshsinha17 The text is not required without any link but suppose if there will be a link present then it is hard to get the text from `links: [ ]`. Also in the `config.yml.hbs` file the footer is present with field text without any link URL.

niteshsinha17:
> > > he suggested using links: [ ] but if I will use this then it will be hard to get the text for the link.
> > 
> > 
> > @codesankalp Can you please explain why do we need a text without any link in the footer links section?
> 
> @niteshsinha17 The text is not required without any link but suppose if there will be a link present then it is hard to get the text from `links: [ ]`. Also in the `config.yml.hbs` file the footer is present with field text without any link URL.

I think you are right

codesankalp:
> @codesankalp I think it's not an issue only if new links are added manually. It is better than checking of url of all the links even if it have a valid url. What do you think?

I can't get this point. Can you explain to me the format you are suggesting by taking an example of a link?

niteshsinha17:
@codesankalp Lets wrap this now.
> Also in the config.yml.hbs file the footer is present with field text without any link URL.

I don't know what was the intention of doing this. But t don't thing there any point of having a link in the footer without a link.

>I can't get this point. Can you explain to me the format you are suggesting by taking an example of a link?

Its not so confusion part.  You know we can edit config file of an org any time. We are just having empty array initially to avoid that warning that also means having no link initially. We can change it when we want to add links in the footer for any org.

nemesisdesign (commented):
ok, good, I think we should do the smae for header, can you proceed please?

nemesisdesign (approved):

PR #216, opened 2021-05-03
[wifi-login-pages] Added logout without breaking auto-login #171

### LINKED ISSUE #171: [wifi-login-pages] Logout without breaking auto-login

codesankalp:
Added support for custom logout when user want to auto-login next time.

Closes #171

### DIFF
--- a/client/components/status/__snapshots__/status.test.js.snap
+++ b/client/components/status/__snapshots__/status.test.js.snap
@@ -3,6 +3,44 @@
 exports[`<Status /> rendering should render correctly 1`] = `
 <React.Fragment>
   <div
+    className="logout-modal"
+  >
+    <div
+      className="logout-modal-container"
+    >
+      <button
+        className="logout-modal-close-btn"
+        onClick={[Function]}
+        type="button"
+      >
+        X
+      </button>
+      <p>
+        Do you want to automatically log in the next time?
+      </p>
+      <ul
+        className="modal-buttons"
+      >
+        <li>
+          <button
+            onClick={[Function]}
+            type="button"
+          >
+            Yes
+          </button>
+        </li>
+        <li>
+          <button
+            onClick={[Function]}
+            type="button"
+          >
+            No
+          </button>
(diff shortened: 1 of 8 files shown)


codesankalp:

nemesisdesign (changes requested):
Good progress @codesankalp, here's what I found during testing, see below.

- the yes/no buttons styles are too different from the rest of the application, try to reuse similar styles or make new ones which are similar
- the yes/no buttons should be clickable
- If I click on yes, in the new page I can see the links to "Status", "Change password" and so on, therefore it means the application thinks I'm still logged in, this shouldn't be this way, the user should be flagged as logged out and should not be able to access the pages which require authentication

Let me clarify the goal of this change: users should be logged out as usual, with the exception that when they click on "yes", the session information is not erased so they're recognized when they'll open to the login page again at a next visit.

Please also ensure (and write a test for it) that the captive portal log out is always called when the log out action is performed.

codesankalp:
Closing this due to https://github.com/openwisp/openwisp-wifi-login-pages/pull/219

PR #219, opened 2021-05-06
[feature] Logout without breaking auto login #171

### LINKED ISSUE #171: [wifi-login-pages] Logout without breaking auto-login

codesankalp:
Logout action is performed without removing cookies and sessions

when auto login is set for the next time.

Closes #171


nemesisdesign (commented):
I pushed some improvements, please check che code changes, test them and verify all is ok, let me know if there's any issue.

codesankalp:
> I pushed some improvements, please check che code changes, test them and verify all is ok, let me know if there's any issue.

Yes, tested them and the style looks much better than the previous one.
It is working fine.

nemesisdesign (changes requested):
Thanks @codesankalp, while testing I found a corner case to handle:

- log in
- then log out (clicking on yes)
- now stop the development server of openwisp-radius
- click log in again

There's an error (rightly so) and we're taken back to the same page, can you make sure that if an error happens while logging in again the user, we redirect to the login page?

nemesisdesign (changes requested):
I got mixed results during testing.

### Going to login page again does not auto log in

Try this:

- log in
- log out, click yes in the modal
- now go back to sign in without clicking on "log me in again"

Following this path the user does not get logged back in again.
Why is that?

Please read the issue (#171) description again:

> When a user logs out, once they go back to the log in page, they will not be auto logged in if they enabled the remember me checkbox.
> 
> We need to find a way to make the auto-login feature work when the remember-me checkbox is enabled, so that when they connect to the WiFi again, they'll be automatically logged in when they open the page.

The current implementation does not meet the goal: when the user is presented the captive page again, they will not be automatically logged in. They do only if they click on that button ("log me in again"), but it's pointless, because users will close the page.

All this work we are doing is to allow users to log in automatically without having to insert username and password each time, **is the goal clear**?

### Testing with a captive portal is giving me different results

Adding the captive portal into the mix gives me totally and different and incorrect results.
In the next days I will work on adding a couple debug views in opewnsip-radius which we can use in this app.
I will try to replicate the same issues and report here.

**Update**: during testing, this condition is not triggered because of these lines:

https://github.com/openwisp/openwisp-wifi-login-pages/blob/9133c58eaef774cde9e41c0b4c299f0dffe77a16/client/components/status/status.js#L185-L186

If you temporarily disable that if, and let the code inside the if run, you will be able to replicate the same behavior I see.

codesankalp:
> Thanks @codesankalp, while testing I found a corner case to handle:
> 
> * log in
> * then log out (clicking on yes)
> * now stop the development server of openwisp-radius
> * click log in again
> 
> There's an error (rightly so) and we're taken back to the same page, can you make sure that if an error happens while logging in again the user, we redirect to the login page?

I fixed this.
Will push the commit soon.

nemesisdesign (changes requested):
I think the lazy loading is causing issues during captive portal logout but it's hard to replicate unless you set up a captive portal.

Here's my understanding of what is happening:

- log out happens, access to the network is shut down for some instants by the firewall
- even though the firewall should allow access to the domain where our app is hosted, for a few seconds this doesn't happen
- this causes the download of the JS chunk which has the code for "logged out successfully" to fail downloading
- therefore the page stays blocked loading, sometimes it finally shows up after 10-20 seconds, sometimes seems to not show at all

I tried eliminating Lazy loading for Logout and seems to work.

The issue happens also if I click on "no" to "do you want to login again...", because going back to log in triggers the download of the chunk which contains `Login` so I think we have to give up lazy loading for both `Login` and `Logout` components.

So I suggest to proceed as follows:

- rebase this PR on the latest master and fix conflicts
- disable lazy loading for `Logout`
- find out if there's anything you can do to shrink the size of logout further
- disable also lazy loading for Login component but in a separate commit

With these changes we should be done, I will do some final testing to confirm.

codesankalp:
> find out if there's anything you can do to shrink the size of logout further

It's of 267 B now.

nemesisdesign (commented):
Looks good! I will give it another round of testing asap.

nemesisdesign (approved):

nemesisdesign (approved):

PR #224, opened 2021-05-07
[Fix] Separated webpack analyzer plugin from main plugins #223

### LINKED ISSUE #223: [bug] Webpack analyzer is causing issues
Sometimes building hangs because of webpack analyzer.

Can we avoid to call webpack analyzer during the standard build process?
Can we avoid to open the webpack analyzer webpage after `yarn start`?

We should trigger webpack analyzer explicitly with a command, it should not be part of the standard process.

codesankalp:
Now webpack analyzer can be avoided during the standard build process and yarn start

Fixes #223

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,8 @@
     "format": "prettier --write \"**/*.+(js|jsx|json|css|md)\"",
     "format:check": "prettier -c \"**/*.+(js|jsx|json|css|md)\"",
     "test": "jest",
-    "test-inspect": "node inspect ./node_modules/.bin/jest --runInBand"
+    "test-inspect": "node inspect ./node_modules/.bin/jest --runInBand",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -99,7 +99,8 @@
     "format": "prettier --write \"**/*.+(js|jsx|json|css|md)\"",
     "format:check": "prettier -c \"**/*.+(js|jsx|json|css|md)\"",
     "test": "jest",
-    "test-inspect": "node inspect ./node_modules/.bin/jest --runInBand"
+    "test-inspect": "node inspect ./node_modules/.bin/jest --runInBand",
+    "stats": "STATS=true npx webpack-cli --profile --json --config config/webpack.config.js > compilation-stats.json && webpack-bundle-analyzer dist/stats.json"
```
nemesisdesign, line 103: can you please update the README section which mentions webpack analyzer so that it recommends using this command to see the results of the analysis please?

nemesisdesign (changes requested):
Good job, it works. See my request below.

nemesisdesign (approved):
Thanks!

PR #225, opened 2021-05-11
[feature] Add support for is_verified field

### LINKED ISSUE #172: [wifi-login-pages] Change: add support for new is_verified field
In the SMS verification flow, right now we check the `is_active` field to know if we should send the user to verify their phone number.

We have to switch to the new `is_verified` field once https://github.com/openwisp/openwisp-radius/issues/233 is completed, because `is_active` can be used also to ban users.

codesankalp:
Closes #172


internals/generators/organization/index.js
```diff
@@ -38,6 +38,12 @@ module.exports = {
         return "The organization uuid is required";
       },
     },
+    {
+      type: "confirm",
+      name: "identity_verification",
+      message: "Does this organization supports identity verification?",
+      default: false,
+    },
```
nemesisdesign, line 46: I don't think we need to add this

server/controllers/registration-controller.js
```diff
@@ -39,6 +39,13 @@ const registration = (req, res) => {
         birth_date,
         location,
       };
+      if (
+        settings &&
+        settings.mobile_phone_verification &&
+        settings.identity_verification
+      ) {
+        postData.identity_verification = "mobile";
```
nemesisdesign, line 47: can we simplify?

```javascript
if (settings.mobile_phone_verification) {
  postData.identity_verification = "mobile";
} else {
  postData.identity_verification = null
}
```

org-configurations/default-configuration.yml
```diff
@@ -4,6 +4,7 @@ slug: "default"
 
 settings:
   mobile_phone_verification: false
+  identity_verification: false
```
nemesisdesign, line 7: if we can simplify as suggested in the previous comment, this will not be necessary

internals/generators/organization/config.yml.hbs
```diff
@@ -3,6 +3,7 @@ name: "{{name}}"
 slug: "{{slug}}"
 settings:
   mobile_phone_verification: {{mobile_phone_verification}}
+  identity_verification: {{identity_verification}}
```
nemesisdesign, line 6: if we can simplify as suggested in the previous comment, this will not be necessary

client/actions/dispatchers.js
```diff
@@ -2,6 +2,7 @@ import logoutAction from "./logout";
 import {
   SET_AUTHENTICATION_STATUS,
   SET_MOBILE_PHONE_VERIFICATION_STATUS,
+  SET_USER_BANNED_STATUS,
```
nemesisdesign, line 5: for consistency with the backend, wouldn't it be clearer if we called this `IS_ACTIVE` instead of `BANNED`?

client/actions/dispatchers.js
```diff
@@ -19,3 +20,8 @@ export const logout = (dispatch) => {
     dispatch(logoutAction(cookies, slug));
   };
 };
+export const userBanned = (dispatch) => {
```
nemesisdesign, line 23: I'd call this `setIsActive`

client/components/mobile-phone-verification/mobile-phone-verification.test.js
```diff
@@ -239,7 +239,7 @@ describe("Mobile Phone Token verification: corner cases", () => {
     sessionStorage.clear();
   });
 
-  it("should not proceed if user is already active", async () => {
+  it("should not proceed if user is already verified", async () => {
```
nemesisdesign, line 242: thanks for updating this kind of things, please double check comments and other parts like this, I will double check tomorrow.

nemesisdesign (changes requested):
Great progress @codesankalp, I did a quick test of the basics and it seems to work! :+1:
See my comments below to simplify the code a bit.
Tomorrow or in the next days I'll do more testing.

nemesisdesign (changes requested):
It's working pretty well!

I found a couple of issues during testing.

### 1. Sign up with mobile phone verification off doesn't redirect to status page 

Try this by creating an organization in OpenWISP RADIUS which does not have SMS verification on, same on wifi-login-pages app, create an org YAML file with mobile verification set to false.

Now, if you try to sign up with a dummy account with the master branch of this app, you should be automatically redirected to the status page after signing up.

On this branch instead, I do not get redirected but I have to log in again.
It would have been great to have a test which failed for this! That's why I created https://github.com/openwisp/openwisp-wifi-login-pages/issues/229 (you can add a test to this branch if you want).

### 2. Verify mobile phone quirk

sms-verification-issue

While testing, initially the verify phone token HTTP requests were sent multiple times, causing the quirk shown in the GIF above.

**However, after trying and retrying several times, the issue went away**.

I am not sure if this is a bug or was a temporary glitch in my development environment,
I ask you to please try double check by trying the same steps and let me know.

### Rebasing

I may merge https://github.com/openwisp/openwisp-wifi-login-pages/pull/228, if I do, please rebase on the latest master.

Purhan (commented):
@codesankalp 
First issue: Please make sure to add a parameter `"identity_verification": "mobile"` while making a call to the registration API for orgs that require mobile verification, this gets stored in the backend


https://user-images.githubusercontent.com/10897423/118300054-2546f700-b4ff-11eb-8df9-0949e14808a1.mp4

Second issue:
I temporarily made this field non-mandatory just to check. It seems the Bearer token is not being set properly which is why verification is also failing. The same token worked in the DRF interface. This may only be a problem with my environment so please double check this.



https://user-images.githubusercontent.com/10897423/118301256-7dcac400-b500-11eb-99bf-7836b3285398.mp4

codesankalp:
@nemesisdesign 
Some fixes that were addressed in this commit are:
1. Sign up with mobile phone verification off now redirects to the status page
2. Verify mobile phone quirk - I can't replicate this one from the frontend but I have integrated the `is active` flow with this also. But if multiple requests will be made it will show the `user is already verified`. For this I have a solution to check for validateToken when the status code is 400 to bypass this error. But again it will call validateToken one more time which I think is not required and increase API calls. Also when we make it (validateToken) reusable then I think this will be resolved.
3. Checking for is_active on the status page - I thought to do this because if a user supports auto-login and from the openwisp-radius it is now changed to inactive then there must be check to perform log out for that user.
4. Added is active action in the set organization.

client/components/status/status.js
```diff
@@ -77,6 +83,10 @@ export default class Status extends React.Component {
       }
       const isValid = await this.validateToken();
       const {is_active, is_verified} = this.state;
+      if (!is_active) {
```
nemesisdesign, line 86: if we decide to keep this, please add a comment like:

```
// if the user is being automatically logged in but it's not
// active anymore (eg: has been banned)
// automatically perform log out
```

nemesisdesign (changes requested):
> Some fixes that were addressed in this commit are:
> 1. Sign up with mobile phone verification off now redirects to the status page

Works :+1:

> 2. Verify mobile phone quirk - I can't replicate this one from the frontend but I have integrated the `is active` flow with this also. But if multiple requests will be made it will show the `user is already verified`. For this I have a solution to check for validateToken when the status code is 400 to bypass this error. But again it will call validateToken one more time which I think is not required and increase API calls. Also when we make it (validateToken) reusable then I think this will be resolved.

Yes ok.

> 3. Checking for is_active on the status page - I thought to do this because if a user supports auto-login and from the openwisp-radius it is now changed to inactive then there must be check to perform log out for that user.

After testing again I found another quirk which I believe is happening for this change.

I can replicate it consistently as follows:

- register a user with phone verification
- validate the phone number
- logout
- login again

At this point I am being automatically logged out.

Then if I log in again I can log in.

But if I log out and log in once more time then this happens again.

I tried fiddling with the code and found out that when this happens it's because `state.is_active` is `null`.
Actually if I check the whole state is kind of empty, although not always, it happens only 1 time out of 2.
This is really a weird quirk and I don't think it's caused by your change but it's something that has been there for a while, so we could address this in the next work we'll be doing.

If you can find a good solution now, go ahead, otherwise please create a new issue for this specific
check (logging out auto-logged in users which are not flagged as active any more) and we'll reintroduce it
once the quirks with the token validation and state are fixed.

> 4. Added is active action in the set organization.

I don't understand this, can you please clarify?

nemesisdesign (changes requested):
PS: please also rebase on the latest master.

codesankalp:
> > 1. Added is active action in the set organization.
> 
> I don't understand this, can you please clarify?

The redux state for `isActive` and `verifyMobileNumber` does not remain in the store when we reload the page manually, so to overcome this I have to add this.

This also occurs with `verifyMobileNumber` but since it returns undefined it doesn't give any problem. I have checked it using the redux dev tools.

Still `isActive` is being set in different components that's why I have to do this, But when reusable-validate-token logic is done it will not be a problem anymore.

codesankalp:
> check (logging out auto-logged in users which are not flagged as active anymore) and we'll reintroduce it

I have checked it, It is giving the expected results.

client/components/status/status.js
```diff
@@ -293,7 +283,18 @@ export default class Status extends React.Component {
...
         }
-        this.setState({username, password, is_active, is_verified, userInfo});
+        this.setState(
+          {username, password, is_active, is_verified, userInfo},
+          () => {
+            // if the user is being automatically logged in but it's not
+            // active anymore (eg: has been banned)
+            // automatically perform log out
+            setIsActive(is_active);
+            if (!is_active) {
```
nemesisdesign, line 293: just to be extra sure, I would do `is_active === false`. I'll try that now and if works will push it here.

PR #227, opened 2021-05-12
[change] Increase size of inputs

nemesisdesign:


client/index.css
```diff
@@ -247,5 +257,12 @@ h1 {
   .side-column {
     margin-top: 20px;
   }
-  /* body .header-row-1-inner { padding: 10px 15px } */
+  #remember_me {
+    width: 25px;
+    height: 25px;
+    bottom: -6px;
+  }
```
codesankalp, line 264: https://github.com/openwisp/openwisp-wifi-login-pages/blob/8e32db4481f5bf0657f14687d734568d88989048/client/components/login/index.css#L36-L42

This is also defined here. I think doing it in a single place will be the right option.

nemesisdesign: right, I moved it into the same file, thanks!
https://github.com/openwisp/openwisp-wifi-login-pages/pull/227/commits/b23fbba2599f13c38d966deec313b710491d06d2

PR #228, opened 2021-05-12
[feature] Allow forcing text input for username field in login form #89

### LINKED ISSUE #89: [feature] Feature: allow logging in using username, email or phone number
In most cases, using the email to sign up and log in will be the best option.

When RADIUS is configured to use realms (think of it like a federation of user access databases), the `@` character is the default delimiter which separates the username from the realm, so an email address would break this RADIUS feature.

Therefore, we need to support the possibility of configuring the login to have the `username` field.

We should take into account the SMS verification as well. We may default the username to the phone number if SMS verification is enabled.

nemesisdesign:
Closes #89

### DIFF
--- a/internals/generators/organization/config.yml.hbs
+++ b/internals/generators/organization/config.yml.hbs
@@ -97,7 +97,7 @@ client:
       input_fields:
         phone_number:
           placeholder:
-            en: "enter mobile phone number"
+            en: "mobile phone number"
           label:
             en: "mobile phone number"
           # default international prefix
@@ -110,7 +110,7 @@ client:
           pattern_description:
             en: "only letters, numbers, and @/./+/-/_ characters"
           placeholder:
-            en: "enter username"
+            en: "username"
           label:
             en: "username"
         email:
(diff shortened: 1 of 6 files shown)


internals/generators/organization/config.yml.hbs
```diff
@@ -147,7 +147,7 @@ client:
           pattern_description:
             en: "only letters, numbers, and @/./+/-/_ characters"
           placeholder:
-            en: "enter first name"
+            en: "first name"
```
codesankalp, line 150: Not related to this PR feature i.e. placeholder and label are the same.

codesankalp (approved):
Tested locally by changing each config for login form and it is working.

Shouldn't this be configurable while adding `org` i.e. `yarn add-org`?

internals/generators/organization/config.yml.hbs
nemesisdesign: you're right, but they were kind of mixed before so I made them all the same, I removed "enter" to keep it short so it always fit on small screens.

nemesisdesign:
> Tested locally by changing each config for login form and it is working.
> 
> Shouldn't this be configurable while adding `org` i.e. `yarn add-org`?

I double checked this, but it makes sense only when mobile verification is on and I am not sure if it's possible to add a conditional prompt. Anyway, it's kind of an advanced use case and I think the info in the README should suffice.

PR #242, opened 2021-05-24
[feature] Reusable auth token validation logic

### LINKED ISSUE #100: [wifi-login-app] Reusable auth token validation logic
As the application increase in complexity and identity verification methods are added like the one in #92, we need to check if the user is authenticated and allowed to see a specific page more often. Right now, some of this logic has been duplicated in #92 and is a bit cumbersome to manage, for this reason I started to think about a possible generic and reusable solutions for the current and future use cases.

A possible reusable `validateToken` function would work this way:

- if no auth token found in cookies, clear any session storage and return false
- if auth token in cookies and was not validated yet, proceed with validation
  - if validation fails, logout, ensure cookies and session storage are cleared, return false
  - if validation succeeds, store the returned data in a general state (redux?), possibly save it into sessionStorage to avoid repeating the operation if the user reloads the page, return true
- if token was already validated and previously fetched data, return true

### User data

Data which will be returned from the [openwisp-radius](https://openwisp-radius.readthedocs.io/en/latest/) API after successful validation:

- username
- email
- phone number
- is_active
- verification_method (this still needs to be defined/added)
- radius_token

### Flow

The logic to show a page only if user is authenticated/unauthenticated or to redirect to another page right now  flow is defined in the `OrganizationWrapper` component, but this splits the logic in two parts, on one side we have the actual component, on the other side we have `OrganizationWrapper` which knows when to display the page and when to redirect or perform a different action, I think this makes things more complicated: harder to debug and automatically test.
I think we aim to make `OrganizationWrapper` a simple list of routes and define the flows in the component themselves.

We should then find a way that would allow us to define the following flow, with less repetition and boilerplate as possible:

- **Login**
  - if authenticated: go to status
  - if not authenticated: show
- **Status**
  - if authenticated and active: show
  - if authenticated and inactive: redirect to identity verification (eg: mobile phone verification)
  - if not authenticated: go to login
- **Sign up**
  - if authenticated: go to status
  - if not authenticated: show
- **Verify Mobile Phone**
  - if not authenticated: go to login
  - if authenticated and active: go to status
  - if authenticated and not active and verification method is mobile phone: show
  - if authenticated and not active and verification method is not mobile phone: go to correct verification method component
- **Change Phone Number**
  - if authenticated: show
  - if not authenticated: go to login
- **Change Password**
  - if authenticated: show
  - if not authenticated: go to login
- **Reset password**
  - if authenticated: status
  - if not authenticated: show

### LINKED ISSUE #217: [bug] Validate token endpoint being called unnecessarily
When logging in on the current master branch, I see that the endpoint which validates the user auth token is called 3 times:

Screenshot from 2021-05-03 20-18-47

I am not sure why this is happening but it looks wrong to me.

Moreover, since I am logging in, I am just getting the token from the API so it should not be validated.

We have to ensure the following:

- whatever data which is got after validating the token can be sent also when the token is received after first authentication, so we can avoid having to call validate
- the validate token endpoint should be used only when the user is being recognized from an existing session or an older session and should be only called once

I think this is related to #100.

### LINKED ISSUE #222: [bug] Status page shows phone number validation briefly
When mobile verification is enabled, after logging in, the mobile verification page is shown briefly for an instant even if the user is already verified.

This can be replicated by adding a `sleep(1.5)` in the `post` method of `ValidateAuthTokenView` of openwisp-radius.

This is related to #100 and #217. It could be fixed with a workaround but the best is to modify the system so that the user data is retrieved when logging in and saved in the general state.

### LINKED ISSUE #251: [bug] Username error is showing for inactive user
The username error must not be shown when the user is inactive.

codesankalp:
Closes #100 
Fixes #217
Fixes #222
Fixes #251


codesankalp:
Working on improving tests.

client/components/mobile-phone-change/mobile-phone-change.js
```diff
@@ -272,4 +252,6 @@ MobilePhoneChange.propTypes = {
   cookies: PropTypes.instanceOf(Cookies).isRequired,
   logout: PropTypes.func.isRequired,
   verifyMobileNumber: PropTypes.func.isRequired,
+  userData: PropTypes.object.isRequired,
```
nemesisdesign, line 255: I think we need to define this further, at least the required attributes we need for the code to work

client/components/login/login.js
```diff
@@ -477,6 +479,7 @@ Login.propTypes = {
   authenticate: PropTypes.func.isRequired,
   verifyMobileNumber: PropTypes.func.isRequired,
   setIsActive: PropTypes.func.isRequired,
+  setUserData: PropTypes.func.isRequired,
```
nemesisdesign, line 482: I think that now that we are introducing this, we should be able to get rid of `verifyMobileNumber` and `setIsActive`.
To do that we should make sure that `setUserData` can edit one specific attribute.

client/components/mobile-phone-change/mobile-phone-change.js
```diff
@@ -70,6 +95,7 @@ class MobilePhoneChange extends React.Component {
         this.setState({
           errors: {},
         });
+        setUserData({});
```
nemesisdesign, line 98: this line is not clear

client/components/mobile-phone-change/mobile-phone-change.js
```diff
@@ -35,17 +31,46 @@ class MobilePhoneChange extends React.Component {
...
+      setUserData,
+      userData,
+    );
+    if (isValid) {
+      ({userData} = this.props);
+      const {phone_number} = userData;
+      this.setState({phone_number});
+      verifyMobileNumber(true);
+    } else {
+      logout(cookies, orgSlug);
```
nemesisdesign, line 57: I am not convinced we should add this kind of logic in every component, looks duplicated

client/utils/validateToken.js
```diff
@@ -0,0 +1,49 @@
...
+        logError(
+          response,
+          '"response_code" !== "AUTH_TOKEN_VALIDATION_SUCCESSFUL"',
+        );
+        return false;
+      }
+      setUserData(response.data);
+      return true;
+    } catch (error) {
+      // logout will be handled from HOC
```
nemesisdesign, line 36: what is HOC?

client/utils/validateToken.js
```diff
@@ -0,0 +1,49 @@
...
+      logError(error, genericError);
+      return false;
+    }
+  }
+  // returns user data if already exists
+  else if (userData && Object.keys(userData).length !== 0) {
+    return true;
+  } else {
```
nemesisdesign, line 44: when is this supposed to happen?

client/utils/validateToken.js
codesankalp: Higher order component where this `validateToken` is used.

client/components/mobile-phone-change/mobile-phone-change.js
codesankalp: Actually, on the status page, we have to handle session logout for the captive portal also. That's why I though to handle logout from the component itself.

client/utils/validateToken.js
codesankalp: When the user data is already present i.e.so it will skip API calling for validating token.
To skip multiple API calls for validating token.

client/components/mobile-phone-change/mobile-phone-change.js
nemesisdesign: we can provide a callable which can default to this action, then if any component needs to customize it can pass a more complex callable

client/utils/validateToken.js
nemesisdesign: please add a comment that explains this

nemesisdesign (changes requested):
I see progress but many points in my previous review have not been addressed:
https://github.com/openwisp/openwisp-wifi-login-pages/pull/242#pullrequestreview-667192771

client/components/login/login.test.js
```diff
@@ -264,49 +263,6 @@ describe("<Login /> interactions", () => {
...
-    wrapper.find("form").simulate("submit", event);
-    await tick();
-    expect(handleSubmit).toHaveBeenCalled();
-    const verifyMock = login.props().verifyMobileNumber.mock;
-    expect(verifyMock.calls.length).toBe(1);
-    expect(verifyMock.calls.pop()).toEqual([true]);
-    const authenticateMock = login.props().authenticate.mock;
-    expect(authenticateMock.calls.length).toBe(1);
-    expect(authenticateMock.calls.pop()).toEqual([true]);
-  });
```
nemesisdesign, line 309: why was this test removed? Shouldn't we check that the mobile verification is initiated as expected because `setUserData` is called with `{method: "mobile_phone", is_verified: false}`?

client/components/login/login.test.js
```diff
@@ -385,11 +314,36 @@ describe("<Login /> interactions", () => {
...
     const event = {preventDefault: () => {}};
     await wrapper.instance().handleSubmit(event);
-    const verifyMock = wrapper.instance().props.verifyMobileNumber.mock;
-    expect(verifyMock.calls.length).toBe(0);
     const authenticateMock = wrapper.instance().props.authenticate.mock;
     expect(authenticateMock.calls.length).toBe(0);
-    const setIsActiveMock = wrapper.instance().props.setIsActive.mock;
-    expect(setIsActiveMock.calls.length).toBe(1);
+    const setUserDataMock = wrapper.instance().props.setUserData.mock;
+    expect(setUserDataMock.calls.length).toBe(1);
```
nemesisdesign, line 320: can you check that it gets called with the right arguments please?

client/utils/validateToken.js
```diff
@@ -0,0 +1,66 @@
+import qs from "qs";
```
nemesisdesign, line 1: pay attention, the name of the file is not consistent with the rest of the codebase

nemesisdesign (changes requested):
Looks on the right track, please see my comments below.

nemesisdesign (changes requested):
During functional testing I found out a few issues.

### Captive portal login being called also when page is not reloaded

validate-token-logic

I was testing that the validate token endpoint was not being called when changing pages, which works, that's great!

However, I noticed something which should not happen: the captive portal login form URL should be called only when the user is either logging in or being recognized and logged in back automatically. We should avoid calling the captive portal login URL should not be called if the user is already authenticated in the app.

If you prefer, we can move this to a dedicated issue and work on it as a follow up to this PR. If you like this approach please copy this text (including the gif and create a new issue).

PS: we'll need a test for this change/fix.

### Inactive user glitch

I flagged a user as inactive and the username field is showing an error which shouldn't be there:

Screenshot from 2021-05-28 13-29-58

Please add an automated test which ensures no errors are shown in the form if the user is inactive.

nemesisdesign (changes requested):
While taking a closer look at the code, I found out the following:

- `verifyMobileNumber` is still lying around, please do a project wide search and ensure is removed

codesankalp:
> While taking a closer look at the code, I found out the following:
> 
> * `verifyMobileNumber` is still lying around, please do a project wide search and ensure is removed

Ok, will recheck again and remove it.

nemesisdesign (changes requested):
Here's another issue I found out during testing.

Try simulating high network latency by adding these two lines to openwisp-radius:

```diff
diff --git a/openwisp_radius/api/views.py b/openwisp_radius/api/views.py
index 0311e0e..5b6a03f 100644
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -313,11 +313,15 @@ class ValidateAuthTokenView(
         """
         Used to check whether the auth token of a user is valid or not.
         """
+        from time import sleep
+        sleep(5)
```

Then, log in, once you are redirected to the status page, reload.

The expected behavior is that the loading overlay covers the page while the data is loaded.
But this is not happening.

I suggest making sure the logic to do this can be also reused, so we don't need to write it in each component.

codesankalp:
I have solved all the above-mentioned changes.

client/components/login/login.js
```diff
@@ -178,20 +172,15 @@ export default class Login extends React.Component {
     localStorage.setItem("rememberMe", remember_me);
     setLoading(true);
 
-    const handleAuthentication = (
-      needsMobileVerification = false,
-      data = {},
-    ) => {
+    const handleAuthentication = (data) => {
+      setUserData(data);
```
nemesisdesign, line 176: why did you move this here? It seems like an unnecessary change.

client/components/login/login.js
```diff
@@ -206,27 +195,25 @@ export default class Login extends React.Component {
...
         if (
           error.response.status === 401 &&
           settings.mobile_phone_verification &&
           data.is_active
         ) {
-          return handleAuthentication(true);
+          return handleAuthentication(data);
         }
+        userData.is_active = false;
+        setUserData(userData);
```
nemesisdesign, line 210: why can't we simply call this earlier and avoid having it both here and in `handleAuthentication`?

client/components/login/login.js
```diff
@@ -206,27 +195,25 @@ export default class Login extends React.Component {
...
-        if (res.data) setIsActive(res.data.is_active);
-        return handleAuthentication(
-          settings.mobile_phone_verification,
-          res.data,
-        );
+        return handleAuthentication(res.data);
       })
       .catch((error) => {
         const {data} = error.response;
-        setIsActive(data.is_active);
```
nemesisdesign, line 217: here, it could be called here. Why did you opt for not doing this and instead call `setUserData()` both in `handleAuthentication` and then below? Isn't this redundant?

Keep in mind redundancy like this is what we have to get rid of. We need to simplify the code and make it easier to deal with. These minor details are not helping us in achieving the goal.

I want to hear your justification for such code changes before deciding what to do with them.

client/components/login/login.js
```diff
@@ -206,27 +195,25 @@ export default class Login extends React.Component {
...
         if (
           error.response.status === 401 &&
           settings.mobile_phone_verification &&
           data.is_active
         ) {
-          return handleAuthentication(true);
+          return handleAuthentication(data);
         }
+        userData.is_active = false;
```
nemesisdesign, line 209: this is definitely very weird. Why do you need to do this? It's really confusing. Aren't we getting the response from the server? Why would we need to override it?

client/components/login/login.js
```diff
@@ -206,27 +195,25 @@ export default class Login extends React.Component {
...
+        setUserData(userData);
         const errorText = data.is_active
           ? getErrorText(error, loginError)
           : getErrorText(error, userInactiveError);
         logError(error, errorText);
         toast.error(errorText);
+        if (data.is_active === false) data.username = "";
```
nemesisdesign, line 216: why is this needed? What problem does it solve?
If the user is inactive, won't they just be logged out? Then why would we need to set the username as empty string?

client/utils/validate-token.js
```diff
@@ -0,0 +1,66 @@
...
+    } catch (error) {
+      handleLogout(logout, cookies, orgSlug, setUserData);
+      logError(error, genericError);
+      return false;
+    }
+  }
+  // returns true if user data exists and skips calling the API
+  else if (token && userData && Object.keys(userData).length > 0) {
+    return true;
+  } else {
```
nemesisdesign, line 61: when does this case happen? Can you please add a comment?

nemesisdesign (changes requested):
There is a conflict after merging #245. Can you please resolve it?

While checking out those conflicts more closely I found a few key lines which are confusing to me and I really recommend clearing out any confusion and simplify the code as much as possible, please see my comments below for more information.

I will also shortly do another round of manual testing.

nemesisdesign (changes requested):
Result of another round of testing: we're almost there!
**I found one key issue which we need to address with utmost care.** Please read on.

> ### Captive portal login being called also when page is not reloaded
> validate-token-logic
> 
> I was testing that the validate token endpoint was not being called when changing pages, which works, that's great!
> 
> However, I noticed something which should not happen: the captive portal login form URL should be called only when the user is either logging in or being recognized and logged in back automatically. We should avoid calling the captive portal login URL should not be called if the user is already authenticated in the app.
> 
> If you prefer, we can move this to a dedicated issue and work on it as a follow up to this PR. If you like this approach please copy this text (including the gif and create a new issue).

The problem described above is solved but I notice that in some cases it seems the captive portal login is not called when it should be.

**Premise**: I just noticed my previous review had typos, sorry for that.
This may have contributed to some ambiguity.
So let me rephrase this correctly:

We should perform captive portal login in these cases:

- the user has performed successful login from login component
- the user is recognized from the cookie and is logged in again automatically
- the user clicks on "log in again" after logging out and clicking on "yes" in the modal which asks "Do you want to automatically log in the next time?"

I am testing whether the captive portal login and logout are called correctly and the result do not seem to be 100% consistent.

Captive portal logout seems pretty much always called, which is good.

The fact is that captive portal login doesn't seem to be called when it would have to be.

Here is how I can replicate it:

- open the login page as unauthenticated
- log in

Expected result: captive portal login URL is called (when configured).
Actual result: captive portal login URL is not called.

If I hit reload in the status page I see the captive portal login being called, which is good.
The entire sequence is shown in the animated GIF below.

captive-portal-login-issue

However, the fact that the captive portal login URL does not get called is a big deal because it means the app will fail to do what is designed to do and the fact that no test is failing make me terribly worried that we may fail at this inadvertently if we break it again while doing other changes.

This is not your fault, because I know you are not aware of these important details and that's why I am here to guide you and make you aware of it.

**For the reasons above, it is imperative to add a failing test for this specific case, we need to protect ourselves from the possibility that this app may fail at its core job.**

Please also manually test the other cases described above (where I wrote `We should perform captive portal login in these cases:`).
I also recommend testing them again after you fix the main issue to see if the fix solves also the other cases, if multiple fixes are required, ensure there's an automated test for each (hopefully this is not needed).

**BTW**: please take a look at this issue https://github.com/openwisp/openwisp-wifi-login-pages/issues/250, which once solved
will help you and everyone working on this app to take advantage of the
(captive portal mock views I added recently to OpenWISP RADIUS)[https://openwisp-radius.readthedocs.io/en/latest/developer/captive_portal_mock.html].
 
> PS: we'll need a test for this change/fix.

Can you point out to me the test you added for this?

> ### Inactive user glitch
> I flagged a user as inactive and the username field is showing an error which shouldn't be there:
> 
> Screenshot from 2021-05-28 13-29-58

This is fixed :+1:.

> Please add an automated test which ensures no errors are shown in the form if the user is inactive.

Can you please point out the automated test to me?

client/components/login/login.test.js
```diff
@@ -361,11 +361,17 @@ describe("<Login /> interactions", () => {
...
     const authenticateMock = wrapper.instance().props.authenticate.mock;
     expect(authenticateMock.calls.length).toBe(0);
     const setUserDataMock = wrapper.instance().props.setUserData.mock;
     expect(setUserDataMock.calls.length).toBe(1);
     expect(setUserDataMock.calls.pop()).toEqual([{is_active: false}]);
+    expect(wrapper.instance().state.errors).toEqual({
+      username: "",
+      password: "",
+    });
+    expect(spyToast).toHaveBeenCalled();
```
codesankalp, line 374: I added a test for username error in this test case when user is not active.

client/components/status/status.test.js
```diff
@@ -432,6 +432,29 @@ describe("<Status /> interactions", () => {
...
+    wrapper = shallow(<Status {...props} />, {
+      context: {setLoading: jest.fn()},
+      disableLifecycleMethods: true,
+    });
+    const spyFn = jest.fn();
+    wrapper.instance().loginFormRef.current = {submit: spyFn};
+    wrapper.instance().componentDidMount();
+    await tick();
+    expect(spyFn.mock.calls.length).toBe(0);
+  });
```
codesankalp, line 458: This is the test for captive login I have added previously.

client/components/login/login.js
```diff
@@ -177,6 +177,7 @@ export default class Login extends React.Component {
       if (!remember_me)
         sessionStorage.setItem(`${orgSlug}_auth_token`, data.key);
       authenticate(true);
+      sessionStorage.setItem(`${orgSlug}_just_authenticated`, true);
```
nemesisdesign, line 179: I see that in the same method `componentDidMount` you add this entry to `sessionStorage` and then remove it straightaway, so I wonder why use `sessionStorage` then and not use any other solution based on pure JS?

In some corner cases `sessionStorage` may fail or may not be implemented properly (think about mini browsers, android web views, only God knows on what kind of browsers this will be run), why do we want to depend on something that is optional and can fail if this feature is the key that must not fail?

I recommend avoiding session storage here and use pure JS (either react state, a global variable, whatever you like).

client/components/status/status.test.js
```diff
@@ -456,6 +456,31 @@ describe("<Status /> interactions", () => {
...
+    sessionStorage.setItem(`${props.orgSlug}_just_authenticated`, true);
+    wrapper = shallow(<Status {...props} />, {
+      context: {setLoading: jest.fn()},
+      disableLifecycleMethods: true,
+    });
+    const spyFn = jest.fn();
+    wrapper.instance().loginFormRef.current = {submit: spyFn};
+    wrapper.instance().componentDidMount();
+    await tick();
+    expect(spyFn.mock.calls.length).toBe(1);
```
nemesisdesign, line 482: this test doesn't fail without the rest of the changes in this commit. Did you know this?

We must ensure tests fail with a clear message without the changes we introduce. Tests are the canary in the coal mine, please ensure the canary will save a human's life or their death will be pointless :joy: 

Therefore, please make sure:

1. the test fails without your changes
2. the test fails if you break the captive portal login feature on purpose by removing key parts and ensuring with manual testing the feature is broken

Without these 2 points, this code is pointless and we're just adding bloat to an already bloated codebase, please keep these suggestions in mind for any future tests you will be writing.

client/components/status/status.test.js
```diff
@@ -456,6 +456,31 @@ describe("<Status /> interactions", () => {
     await tick();
     expect(spyFn.mock.calls.length).toBe(0);
   });
+  it("should submit loginFormRef after login", async () => {
```
nemesisdesign, line 459: please rename this to something more explicit, eg: `should perform captive portal login (submit loginFormRef)`

client/components/status/status.test.js
```diff
@@ -456,6 +456,31 @@ describe("<Status /> interactions", () => {
...
+      email: "tester@tester.com",
+      is_active: true,
+      is_verified: true,
+      phone_number: "+237672279436",
+    };
+    expect(true).toBe(true);
+    sessionStorage.setItem(`${props.orgSlug}_just_authenticated`, true);
+    wrapper = shallow(<Status {...props} />, {
+      context: {setLoading: jest.fn()},
+      disableLifecycleMethods: true,
```
nemesisdesign, line 476: I think this line should be removed so the `componentDidMount` method should be called automatically without the need of us calling it here.

client/components/status/status.js
```diff
@@ -133,9 +135,10 @@ export default class Status extends React.Component {
             this.loginFormRef &&
             this.loginFormRef.current &&
             justAuthenticated
-          )
+          ) {
             this.loginFormRef.current.submit();
-
+            sessionStorage.removeItem(`${orgSlug}_just_authenticated`);
```
nemesisdesign, line 140: this test does fail without the fixes but the message shown (`should store token in sessionStorage when remember me is unchecked and rememberMe in localstorage`) doesn't reflect the real issue.
That's why I suggest ensuring the other test is the one that should alert us if anything goes wrong with the captive portal login feature.

nemesisdesign (changes requested):
There's progress but we're not at "professional efficiency" yet, please see my comments below.

client/components/login/login.js
codesankalp: I have to use sessionStorage because I want to get the status of just_authenticated from the login component. Another possible solution is using redux but I thought it would be too long for this solution. Should I implement this using redux?

client/components/status/status.js
```diff
@@ -90,7 +90,9 @@ export default class Status extends React.Component {
       setLoading(false);
       if (isValid) {
         const justAuthenticated =
-          Object.keys(userData).length <= Object.keys(initialState).length;
+          sessionStorage.getItem(`${orgSlug}_just_authenticated`) === "true"
+            ? true
+            : Object.keys(userData).length <= Object.keys(initialState).length;
```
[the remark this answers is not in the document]
codesankalp: It works.
I added the condition check for session storage because I want to know if a user is just authenticated or not i.e. if the user came after login then just authenticated will be true in the status page.

client/components/status/status.test.js
```diff
@@ -464,6 +434,53 @@ describe("<Status /> interactions", () => {
       "4e:ed:11:2b:17:ae",
     );
   });
+  it("should not submit loginFormRef if user is already authenticated", async () => {
```
nemesisdesign, line 437: why didn't you update this test description too? Please avoid the need of me telling you everything, If I ask you to make descriptions more explicit in one place, I'm assuming you're smart enough to understand you have to maintain consistency and make it explicit everywhere. Make this description consistent with the one below.

nemesisdesign (changes requested):
Great progress @codesankalp, this is almost ready, please follow the suggestions below and we'll be able to merge this.

- follow the code suggestion below regarding a specific test description before anything else
- squash commits so that we end up with one commit for each related issue
- please ensure each commit follows the commit message guidelines regarding referencing and closing issues:
https://openwisp.io/docs/developer/contributing.html#commit-message-style-guidelines

nemesisdesign (approved):
Ready! :+1:

PR #243, opened 2021-05-24
[Fix] Changed toast for login error when connection refused

### LINKED ISSUE #241: [bug] Your account has been deactivated, but it's not true
Try to log in with the development server of OpenWISP RADIUS stopped.

Screenshot from 2021-05-21 17-43-09

I see the "your account has been deactivated..." toast error notification, but this is wrong.

The message should be the generic error, as it was before (shown in the following screenshot):

Screenshot from 2021-05-21 17-42-24

@codesankalp can you produce a failing test for this bug?

codesankalp:
Fixes #241

### DIFF
--- a/client/components/login/login.js
+++ b/client/components/login/login.js
@@ -222,9 +222,10 @@ export default class Login extends React.Component {
         ) {
           return handleAuthentication(true);
         }
-        const errorText = data.is_active
-          ? getErrorText(error, loginError)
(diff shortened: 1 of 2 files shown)


nemesisdesign:
Please open the PR from a branch of this repository.
Also reply to the original question in the issue: can you replicate this issue in tests? We need to add regression tests for bugs otherwise we risk the same bugs happening again in the future.

PR #244, opened 2021-05-25
[feature] Added server side logging

codesankalp:
Closes #82

Tasks done:

- Added Winston for logging
- Integrated morgan with Winston to log HTTP requests
- Removed checking for status code to handle timeout error and connection refused error.

PR #245, opened 2021-05-25
[Fix] Changed toast for login error when connection refused

### LINKED ISSUE #241: [bug] Your account has been deactivated, but it's not true

codesankalp:
Closes #241

### DIFF
--- a/client/components/login/login.test.js
+++ b/client/components/login/login.test.js
@@ -3,7 +3,7 @@ import axios from "axios";
 import {shallow, mount} from "enzyme";
 import React from "react";
 import ShallowRenderer from "react-test-renderer/shallow";
-import {toast} from "react-toastify";
+import * as dependency from "react-toastify";
 import PropTypes from "prop-types";
 import {Provider} from "react-redux";
 import {Router} from "react-router-dom";
@@ -206,7 +206,7 @@ describe("<Login /> interactions", () => {
         return Promise.resolve();
       });
     const event = {preventDefault: () => {}};
-    const spyToast = jest.spyOn(toast, "error");
+    const spyToast = jest.spyOn(dependency.toast, "error");
 
     return wrapper
       .instance()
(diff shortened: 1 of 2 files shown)


nemesisdesign (changes requested):
Can you please add a failing test for this?

codesankalp:
> Can you please add a failing test for this?

yes

codesankalp:
Now after adding the commit in #244 the toast content is now internal server error.

client/components/login/login.test.js
```diff
@@ -392,4 +392,56 @@ describe("<Login /> interactions", () => {
...
+    };
+    return wrapper
+      .instance()
+      .handleSubmit(event)
+      .then(() => {
+        expect(wrapper.instance().props.authenticate.mock.calls.length).toBe(0);
+        expect(lastConsoleOutuput).not.toBe(null);
+        expect(errorMethod).toHaveBeenCalled();
+        expect(errorMethod).toBeCalledWith("Internal server error");
+        lastConsoleOutuput = null;
```
nemesisdesign, line 420: let's use `afterEach` to do this kind of operation, which is cleaner and more reliable (this line may not be executed if the test fails).
https://github.com/openwisp/openwisp-wifi-login-pages/blob/ceba2a78d9b733c7473e7962ea641351dd5741fa/client/components/registration/registration.test.js#L76-L78

client/components/login/login.test.js
```diff
@@ -392,4 +392,56 @@ describe("<Login /> interactions", () => {
...
+    dependency.toast = {
+      error: errorMethod,
+    };
+    wrapper = shallow(<Login {...props} />, {context: loadingContextValue});
+    await wrapper.instance().handleSubmit(event);
+    expect(wrapper.instance().props.authenticate.mock.calls.length).toBe(0);
+    expect(lastConsoleOutuput).not.toBe(null);
+    expect(errorMethod).toHaveBeenCalled();
+    expect(errorMethod).toBeCalledWith("Login error occurred.");
+    lastConsoleOutuput = null;
```
nemesisdesign, line 445: let's use `afterEach` to do this kind of operation, which is cleaner and more reliable (this line may not be executed if the test fails).
https://github.com/openwisp/openwisp-wifi-login-pages/blob/ceba2a78d9b733c7473e7962ea641351dd5741fa/client/components/registration/registration.test.js#L76-L78

client/components/login/login.test.js
```diff
@@ -392,4 +392,56 @@ describe("<Login /> interactions", () => {
     const setIsActiveMock = wrapper.instance().props.setIsActive.mock;
     expect(setIsActiveMock.calls.length).toBe(1);
   });
+  it("should show error toast when server error", () => {
```
nemesisdesign, line 395: This test does not fail without the fix. Were you aware of it? What is the purpose of this test?

nemesisdesign (changes requested):
Almost there, please see comments below.

client/components/login/login.test.js
codesankalp: I wrote it because when the internal server error is from the openwisp-radius then it shows an internal server error toast.
but when the API shows connection refused or timeout the error is generic i.e. login error occurred.
So I thought to add both tests.
Should I remove the internal server error one?

nemesisdesign: It's ok

nemesisdesign (approved):

PR #259, opened 2021-06-02
[tests] Mocks support added for captive portal login and logout

### LINKED ISSUE #250: [tests] Add support for openwisp-radius captive portal mock views in dev env
I added a couple of URLs in the test environment of openwisp-radius which simulate the actions which are usually performed when captive portal login and logout happens (the creation or closing of the radius session): https://github.com/openwisp/openwisp-radius/pull/264.

It would be very useful to configure the default configuration of the development environment of this repo to use those URLs by default (only dev env, no need to change the config generator) and to add a few lines in the README to explain this and link the new documentation page which is added to the openwisp-radius docs regarding these mock views.

### LINKED ISSUE #257: [bug] Logout is performed when calling `fetchMoreSessions` with incorrect page
To reproduce this bug:
1. Try to log in. (make sure captive portal login also get performed)
2. Now reloading or scrolling down the page calls `getUserRadiusSessions` with an incorrect page number.
3. Automatically logout will be performed because calling `getUserRadiusSessions` returns 404 for the page that does not exist with the response: `{"detail":"Invalid page."}`

**Bug**: Logout is performed for incorrect page.
**Expected Behaviour**: Infinite loader should hide when page limit is reached.

a-2021-06-07_15 01 12

codesankalp:
Closes #250 
Fixes #257

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,6 +341,15 @@ There are certain environment variables used to configure server logging. The de
 
 All the **HTTP requests** get logged by default in the console during development.
 
+#### Mocking captive portal login and logout
+
+The captive portal login and logout operations can be mocked by using the endpoints
+mentioned in [openwisp-radius captive portal mock docs](https://openwisp-radius.readthedocs.io/en/latest/developer/captive_portal_mock.html).
+
+These URLs from OpenWISP RADIUS will be used by default in the development environment.
+The captive portal login and logout URLs and their parameters can be changed by
+editing the YAML configuration file of the respective organization.
+
 ### License
 
 See [LICENSE](https://github.com/openwisp/openwisp-wifi-login-pages/blob/master/LICENSE).
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -42,7 +42,7 @@ export default class Status extends React.Component {
       loggedOut: false,
       userInfo: {},
       currentPage: 1,
-      hasMoreSessions: true,
+      hasMoreSessions: false,
       intervalId: null,
       screenWidth: window.innerWidth,
       loadSpinner: true,
@@ -175,6 +175,7 @@ export default class Status extends React.Component {
         url,
         params,
       });
+      const {headers} = response;
       if (params.is_open) {
         options.activeSessions = response.data;
         options.sessionsToLogout = response.data;
@@ -183,12 +184,8 @@ export default class Status extends React.Component {
         options.pastSessions = pastSessions.concat(response.data);
         options.currentPage = params.page;
       }
-      if (
-        "link" in response.headers &&
-        !response.headers.link.includes("next")
-      ) {
-        options.hasMoreSessions = false;
-      }
+      options.hasMoreSessions =
+        "link" in headers && headers.link.includes("next");
       this.setState(options);
     } catch (error) {
       logout(cookies, orgSlug);
--- a/client/components/status/status.test.js
+++ b/client/components/status/status.test.js
@@ -691,4 +691,28 @@ describe("<Status /> interactions", () => {
     wrapper.find(".modal-buttons button:last-child").simulate("click", {});
     expect(handleLogout).toHaveBeenCalledWith(false);
   });
+  it("should set hasMoreSessions to false if link is not in response headers", async () => {
+    axios.mockImplementationOnce(() => {
+      return Promise.resolve({
+        status: 200,
+        statusText: "OK",
+        data: [
+          {
+            session_id: 1,
+            start_time: "2020-09-08T00:22:28-04:00",
+            stop_time: "2020-09-08T00:22:29-04:00",
+          },
+        ],
+        headers: {},
+      });
+    });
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -341,6 +341,11 @@ There are certain environment variables used to configure server logging. The de
 
 All the **HTTP requests** get logged by default in the console during development.
 
+#### Mocking captive portal login and logout
+
+This mock is only available when the website is running in development environment. Webpack mode decides the `NODE_ENV` for this mock.
+The endpoints for mocking captive portal login and logout are mentioned [here](https://openwisp-radius.readthedocs.io/en/latest/developer/captive_portal_mock.html).
```
nemesisdesign, line 347: avoid links with "here": https://www.w3.org/QA/Tips/noClickHere, use the title of the page as the link text please.

Also avoid long lines, break the lines at about 70 characters (just look at the lines above and try to follow the same line length), this helps us while working on the text in the editor.

client/components/status/status.js
```diff
@@ -722,7 +722,11 @@ export default class Status extends React.Component {
               ref={this.loginFormRef}
               method={captivePortalLoginForm.method || "post"}
               id="cp-login-form"
-              action={captivePortalLoginForm.action || ""}
+              action={
+                process.env.NODE_ENV !== "development"
+                  ? captivePortalLoginForm.action || ""
+                  : "http://localhost:8000/captive-portal-mock/login/"
```
nemesisdesign, line 728: This would mean that if I want to test a real captive portal in a development environment I won't be able to do it.

I suggest just adding these as default values in the default config:

- https://github.com/openwisp/openwisp-wifi-login-pages/blob/master/org-configurations/default-configuration.yml#L421-L442
- https://github.com/openwisp/openwisp-wifi-login-pages/blob/master/internals/generators/organization/index.js#L59-L81

nemesisdesign (changes requested):

client/components/status/status.js
```diff
@@ -149,8 +150,9 @@ export default class Status extends React.Component {
         options.currentPage = params.page;
       }
       if (
-        "link" in response.headers &&
-        !response.headers.link.includes("next")
+        ("link" in response.headers &&
+          !response.headers.link.includes("next")) ||
+        !("link" in response.headers)
```
nemesisdesign, line 155: can you explain me what is going on here? Is this a fix for a bug? If yes how to replicate it and why there's no test for it?

codesankalp: It's a fix for a bug. 
When a new user just signups and log in and log out for few times (suppose 2 times) then the past session length is 2 which is less than the pagination length defined in openwisp-radius i.e. 10. So the API doesn't include `link` in headers because there is no next for past sessions. 
But according to the previous condition, it will not set hasMoreSession to `false` which is required to stop infinite loading of past sessions. So it will fetch the next page which results in an error and automatically log out.

I will write a test for this.

README.md
```diff
@@ -341,6 +341,14 @@ There are certain environment variables used to configure server logging. The de
 
 All the **HTTP requests** get logged by default in the console during development.
 
+#### Mocking captive portal login and logout
+
+Captive portal login and logout can be mocked by using the endpoints
```
nemesisdesign, line 346: The captive portal login and logout operations

README.md
```diff
@@ -341,6 +341,14 @@ There are certain environment variables used to configure server logging. The de
...
+mentioned in [openwisp-radius captive portal mock docs](https://openwisp-radius.readthedocs.io/en/latest/developer/captive_portal_mock.html).
+
+By default, it will be used in the development environment. These endpoints can
+be configured by changing them in the respective organization configuration file.
```
nemesisdesign, line 350: Ideas for improvement of this text:

These URLs from OpenWISP RADIUS will be used by default in the development environment.
The captive portal login and logout URLs and their parameters can be changed by editing the YAML configuration files.

nemesisdesign (changes requested):
Thanks for adding the test. I double checked and everything is working.

However, I find it a bit weird that the default value for `hasMoreSessions` is set to `true`, the code for the header check can be simplified. I sent a commit addressing this (db03e1a), please double check everything is working as expected and try to follow the same approach in your future work (do not be afraid to change the code if the new solution is simpler, but always make sure to make these kind of changes in specific commits so we can easily revert if needed).

So was this the fix for https://github.com/openwisp/openwisp-wifi-login-pages/issues/257?
If yes we need to add the specific commit message text to reference and close the issue.

Once you finish double checking and think the code can be merged, please proceed to squash the current commits so that we end up with 2 commits, one for the captive portal mock and the other for issue #257.

nemesisdesign (commented):
I'd change the commit messages to:

- [tests] Added OpenWISP RADIUS captive portal login and logout mock views #250
- [fix] Fixed incorect logout when sessions do not paginate #257

nemesisdesign (approved):

PR #261, opened 2021-06-03
[feature] Added selenium tests for login and signup

### LINKED ISSUE #213: [ci] Implement solution to quickly merge dependabot PRs
Dependabot is great but it happened recently that I merged some PRs without manual testing and the app broke (couldn't start) because of issues not found during automated tests.

The tests are headless and do not run in the browser, so maybe if we start to add a very basic test with selenium or any other browser based testing tool we should be able to understand if the app is at least starting, right?

This would allow us then to merge dependabot PRs quickly. 

A basic initial set of browser tests could be:

- open the login page and ensure the result matches what we expect and there's no JS errors
- open the sign up page and ensure the result matches what we expect and there's no JS errors

### LINKED ISSUE #229: [tests] Add a test for successful sign up result
A successful sign up should redirect to the status page. This works in master, but wasn't working in https://github.com/openwisp/openwisp-wifi-login-pages/pull/225 and no test was failing.

We need a test for this! Actually, two tests: one for simple email signup and one for mobile phone verification.

codesankalp:
Closes #213 
Closes #229

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,15 +113,26 @@ try to run the tests with the following command:
 yarn test # headless tests
 ```
 
-It is also possible to run browser based tests, before trying this
-please install [gecko driver](https://github.com/mozilla/geckodriver/releases/) and then run:
+##### Browser based tests
+
+Prerequisites for running browser-based tests:
+
+1. [Gecko driver](https://github.com/mozilla/geckodriver/releases/) needs to be installed.
+2. Running instances of openwisp-radius and
+   openwisp-wifi-login-pages is required.
+3. `OPENWIPS_RADIUS_PATH` environment variable is needed to initialize
+   data while running browser test. This can be set using the following command:
+   ```
+   export OPENWISP_RADIUS_PATH=< PATH TO OPENWISP RADIUS DIRECTORY >
+   ```
+4. If a virtual environment is used to run openwisp-radius then
+   this needs to be activated before running browser tests.
+
+After doing all the prerequisites, run browser based tests using the following command:
 
 ```
 yarn browser-test
 ```
-
-**NOTE**: Browser tests needs running instances of openwisp-radius
-and openwisp-wifi-login-pages.
 
 #### Setup
 
--- a/README.md
+++ b/README.md
@@ -110,8 +110,18 @@ To verify all the dependencies were successfully installed,
 try to run the tests with the following command:
 
 ```
-yarn test
+yarn test # headless tests
 ```
+
+It is also possible to run browser based tests, before trying this
+please install [gecko driver](https://github.com/mozilla/geckodriver/releases/) and then run:
+
+```
+yarn browser-test
+```
+
+**NOTE**: Selenium tests needs running instances of openwisp-radius
+and openwisp-wifi-login-pages.
 
 #### Setup
 
@@ -155,17 +165,20 @@ you can use the default sample captive portal login and logout URLs.
 List of yarn commands:
 
 ```
-$ yarn start      # Run the app (runs both, client and server)
-$ yarn setup      # Discover Organization configs and generate config.json and asset directories
-$ yarn add-org    # Add new Organization configuration
-$ yarn build      # Build the app
-$ yarn server     # Run server
-$ yarn client     # Run client
-$ yarn coveralls  # Run coveralls
-$ yarn lint       # Run ESLint
-$ yarn lint:fix   # Run ESLint with automatically fix problems option
-$ yarn test       # Run tests
-$ yarn -- -u      # Update Jest Snapshots
+$ yarn start         # Run the app (runs both, client and server)
+$ yarn setup         # Discover Organization configs and generate config.json and asset directories
+$ yarn add-org       # Add new Organization configuration
+$ yarn build         # Build the app
+$ yarn server        # Run server
+$ yarn client        # Run client
+$ yarn coveralls     # Run coveralls
+$ yarn coverage      # Run tests and generate coverage files
+$ yarn lint          # Run ESLint
+$ yarn lint:fix      # Run ESLint with automatically fix problems option
+$ yarn format        # Run formatters to format the code
+$ yarn test          # Run tests
+$ yarn selenium-test # Run browser based selenium tests
+$ yarn -- -u         # Update Jest Snapshots
 ```
 
 #### Using custom ports
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Prerequisites for running browser-based tests:
    ```
 4. If a virtual environment is used to run openwisp-radius then
    this needs to be activated before running browser tests.
-5. In the test environment of openwisp-radius, the ``default`` organization
+5. In the test environment of openwisp-radius, the `default` organization
    must be present or alternatively
 
 After doing all the prerequisites, run browser based tests using the following command:
--- a/README.md
+++ b/README.md
@@ -118,12 +118,12 @@ yarn test # headless tests
 Prerequisites for running browser-based tests:
 
 1. [Gecko driver](https://github.com/mozilla/geckodriver/releases/) needs to be installed.
-2. Running instances of openwisp-radius and
+2. having running instances of openwisp-radius and
    openwisp-wifi-login-pages is required.
-3. `OPENWIPS_RADIUS_PATH` environment variable is needed to initialize
-   data while running browser test. This can be set using the following command:
+3. `OPENWIPS_RADIUS_PATH` environment variable is needed to setup/tear down the database
+   data needed to run the browser tests. This can be set using the following command:
    ```
-   export OPENWISP_RADIUS_PATH=< PATH TO OPENWISP RADIUS DIRECTORY >
+   export OPENWISP_RADIUS_PATH=<PATH_TO_OPENWISP_RADIUS_DIRECTORY>
    ```
 4. If a virtual environment is used to run openwisp-radius then
    this needs to be activated before running browser tests.
--- a/README.md
+++ b/README.md
@@ -118,8 +118,7 @@ yarn test # headless tests
 Prerequisites for running browser-based tests:
 
 1. [Gecko driver](https://github.com/mozilla/geckodriver/releases/) needs to be installed.
-2. having running instances of openwisp-radius and
-   openwisp-wifi-login-pages is required.
+2. Having running instances of openwisp-radius and openwisp-wifi-login-pages is required.
 3. `OPENWIPS_RADIUS_PATH` environment variable is needed to setup/tear down the database
    data needed to run the browser tests. This can be set using the following command:
    ```
@@ -127,6 +126,8 @@ Prerequisites for running browser-based tests:
    ```
 4. If a virtual environment is used to run openwisp-radius then
    this needs to be activated before running browser tests.
+5. In the test environment of openwisp-radius, the ``default`` organization
+   must be present or alternatively
 
 After doing all the prerequisites, run browser based tests using the following command:
 
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ please install [gecko driver](https://github.com/mozilla/geckodriver/releases/) 
 yarn browser-test
 ```
 
-**NOTE**: Selenium tests needs running instances of openwisp-radius
+**NOTE**: Browser tests needs running instances of openwisp-radius
 and openwisp-wifi-login-pages.
 
 #### Setup
@@ -177,7 +177,7 @@ $ yarn lint          # Run ESLint
 $ yarn lint:fix      # Run ESLint with automatically fix problems option
 $ yarn format        # Run formatters to format the code
 $ yarn test          # Run tests
-$ yarn selenium-test # Run browser based selenium tests
+$ yarn browser-test  # Run browser based selenium tests
 $ yarn -- -u         # Update Jest Snapshots
 ```
 
--- a/browser-test/clear_data.py
+++ b/browser-test/clear_data.py
@@ -1,35 +1,45 @@
 #!/usr/bin/env python
+import json
 import os
 import sys
-import json
-import subprocess
-import re
 
-project_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
-browser_test_path = os.path.join(project_path, 'browser-test')
-test_data_file = open(os.path.join(browser_test_path, 'testData.json'))
-test_data = json.loads(test_data_file.read())
-test_data_file.close()
+import django
+
+
+def load_test_data():
+    project_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+    browser_test_path = os.path.join(project_path, 'browser-test')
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -110,9 +110,13 @@ To verify all the dependencies were successfully installed,
 try to run the tests with the following command:
 
 ```
-yarn test
+yarn test # headless tests
+yarn selenium-test # browser based tests
```
nemesisdesign, line 114: what about calling this command `yarn browser-test`?
If we ever need to change the browser testing engine we won't need to rename the command.

selenium-tests/login.test.js
```diff
@@ -0,0 +1,50 @@
...
+    );
+    username.sendKeys("openwisp@openwisp.org");
+    password.sendKeys("test_password");
+    const submitBtn = await getElementByXPath(
+      driver,
+      "//INPUT[@type='submit']",
+      until,
+      By,
+    );
+    await sleep(1000);
```
nemesisdesign, line 36: There should be a better way to do this. 
Last time I used selenium, I used a method that allowed to wait until a specific element on the page was visible/present, with a timeout. 
See FluentWait in this page: https://www.selenium.dev/documentation/en/webdriver/waits/ for more info.
Applies to all other occurrences of `sleep`.

README.md
nemesisdesign, line 114: The first attempt at running this failed with:

```
The geckodriver executable could not be found on the current PATH. Please download the latest version from https://github.com/mozilla/geckodriver/releases/ and ensure it can be found on your PATH.
```

Please specify that this dependency needs to be installed before running browser based tests. Also separate the two test examples in two different code samples, eg:

```
yarn test

It is also possible to run browser based tests, before trying this please install ..., then run:

yarn browser-test
```

Once installed, the second attempt fails because the data I have in my openwisp radius doesn't match the test.
**We need a way to allow developers to prepare the test env easily, quickly and automatically**.

We could do this with python/shell scripts that are called at the beginning of each browser test.

Eg:

```bash
#!/bin/bash
echo "User.objects.create_user(.....)" | {OPENWISP_RADIUS_PATH}/manage.py shell_plus
```

`OPENWISP_RADIUS_PATH` could be an environment varable.

nemesisdesign (changes requested):
I don't see this test being run in the CI. Why is that?
We need these tests to run in the CI as well, so you'll need to install openwisp-radius and setup some preparation scripts which prepare the data needed for the tests.

Please ensure your commits follow the commit message guidelines for referencing and closing issues:
https://openwisp.io/docs/developer/contributing.html#commit-message-style-guidelines.

browser-test/login.test.js
```diff
@@ -0,0 +1,63 @@
+import {Builder, By, until} from "selenium-webdriver";
+import {getDriver, getElementByXPath, urls, initialData} from "./utils";
+
+const firefox = require("selenium-webdriver/firefox");
+
+describe("Selenium tests for <Login />", () => {
+  let driver;
+  beforeAll(async () => {
+    jest.setTimeout(30000);
+    driver = await getDriver(Builder, new firefox.Options().headless());
```
nemesisdesign, line 10: can you find a way to call the initialization python script from here?
I am not sure if the virtualenv would be used if activated, you could try.

browser-test/login.test.js
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  afterAll(async () => {
+    driver.close();
```
nemesisdesign, line 14: I think we will need a clean up python script to be called when tests are finished, to clear up any data which is created by the tests

nemesisdesign (changes requested):
Looking a lot better!

I think we have to find a way to ensure the data is consitent and tests can be repeated without issues, that is, I think we need initialization and tear down scripts, see below for more info.

nemesisdesign (changes requested):
Looks good!

Please document `OPENWISP_RADIUS_PATH` in the README and explain how to use it to developers which will want to run these tests.
Assume they don't know anything, it could be a future new contributor like you were in the begining, help your future self.

nemesisdesign (commented):

browser-test/login.test.js
```diff
@@ -0,0 +1,71 @@
...
+  initialData,
+  executeCommand,
+} from "./utils";
+
+const firefox = require("selenium-webdriver/firefox");
+
+describe("Selenium tests for <Login />", () => {
+  let driver;
+  beforeAll(async () => {
+    jest.setTimeout(30000);
```
nemesisdesign, line 15: why a timeout of 30 seconds? This operation should not take so long.

browser-test/utils.js
```diff
@@ -0,0 +1,39 @@
+import {exec} from "child_process";
+import testData from "./testData.json";
+
+const waitTime = 20000;
```
nemesisdesign, line 4: why 20 secs? Looks too much to me. My tests were failing because env issues and it took too long to know what's going on.

browser-test/testData.json
```diff
@@ -0,0 +1,7 @@
+{
+  "testuser": {
+    "email": "testuser@openwisp.org",
+    "password": "testuser",
+    "organization": "default"
```
nemesisdesign, line 5: this assumes there's an org named "default" which may not be true, so the script can fail because of this, which is ok if we get a good error message, but no error message is being given.

If you try to delete the default organization you should then see:

```
TimeoutError: Waiting element to become stale
Wait timed out after 5084ms
```

I lowered the timeout to 5 secs, I guess it could be lower as well.

Please make sure that if this case happens a proper error message is printed out, also double check this timeout thing which seems too big, 20/30 seconds is definitely annoying to work with, I put it to 5 secs but I am wondering why do we need such a long timeout in the first place? Ensure it's as low as possible, tests should fail fast if there's an issue like that. The faster they fail the better.

browser-test/clear_data.py
```diff
@@ -0,0 +1,26 @@
...
+project_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+browser_test_path = os.path.join(project_path, 'browser-test')
+test_data_file = open(os.path.join(browser_test_path, 'testData.json'))
+test_data = json.loads(test_data_file.read())
+test_data_file.close()
+
+OPENWISP_RADIUS_PATH = os.getenv('OPENWISP_RADIUS_PATH', '')
+
+if OPENWISP_RADIUS_PATH == '':
+    print('OPENWISP_RADIUS_PATH is needed for initializing data.')
```
nemesisdesign, line 15: If I do not set `$OPENWISP_RADIUS_PATH`, I get:

```
TimeoutError: Waiting element to become stale
Wait timed out after 5084ms
```

(I lowered the timeout to 5 secs as mentioned in my prev comment)

I was expecting to see this error message, but that's not happening, so this part of the code is not working as expected.

Please ensure the error message is shown if the variable is not set.

browser-test/initialize_data.py
```diff
@@ -0,0 +1,30 @@
...
+project_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+browser_test_path = os.path.join(project_path, 'browser-test')
+test_data_file = open(os.path.join(browser_test_path, 'testData.json'))
+test_data = json.loads(test_data_file.read())
+test_data_file.close()
+
+OPENWISP_RADIUS_PATH = os.getenv('OPENWISP_RADIUS_PATH', '')
+
+if OPENWISP_RADIUS_PATH == '':
+    print('OPENWISP_RADIUS_PATH is needed for initializing data.')
```
nemesisdesign, line 15: same here

nemesisdesign (changes requested):
Great progress on this one @codesankalp! :+1: 

:heavy_check_mark: I have stress tested the tests to ensure they fail if the core functionality of login or registration is broken.

:x: I struggled in getting the tests to run because I did some minor mistakes and the huge timeout plus the error messages made it hard to debug, so this needs to be improved, see my comments below .

browser-test/login.test.js
codesankalp: In the CI, if I used 5 seconds as timeout then it's failing because it takes time in initialization and running-test. Can I make it 15 seconds?

browser-test/login.test.js
```diff
@@ -0,0 +1,76 @@
...
+describe("Selenium tests for <Login />", () => {
+  let driver;
+  beforeAll(async () => {
+    jest.setTimeout(10000);
+    await executeCommand("./browser-test/clear_data.py", (message) => {
+      console.log(message);
+    });
+    await executeCommand("./browser-test/initialize_data.py", (message) => {
+      console.log(message);
+    });
```
nemesisdesign, line 21: let's avoid duplicating code like:

```javascript
await executeCommand("./browser-test/initialize_data.py", (message) => {
  console.log(message);
});

```

In every test file, please create 2 reusable test functions: `initializeData` and `clearData`.

browser-test/login.test.js
```diff
@@ -0,0 +1,76 @@
...
+  initialData,
+  executeCommand,
+} from "./utils";
+
+const firefox = require("selenium-webdriver/firefox");
+
+describe("Selenium tests for <Login />", () => {
+  let driver;
+  beforeAll(async () => {
+    jest.setTimeout(10000);
```
nemesisdesign, line 15: again, this is completely unnecessary, please remove it, if you need to get the github actions build to wait, please use the linux `sleep` command in the github actions workflow. Try to see if 5-7 seconds are enough though. If we had to pay for these builds, this would be money thrown out of the window.

browser-test/registration.test.js
```diff
@@ -0,0 +1,73 @@
...
+  initialData,
+  executeCommand,
+} from "./utils";
+
+const firefox = require("selenium-webdriver/firefox");
+
+describe("Selenium tests for <Register />", () => {
+  let driver;
+  beforeAll(async () => {
+    jest.setTimeout(10000);
```
nemesisdesign, line 15: same here for timeout

browser-test/registration.test.js
```diff
@@ -0,0 +1,73 @@
...
+    await executeCommand("./browser-test/clear_data.py", (message) => {
+      console.log(message);
+    });
```
nemesisdesign, line 18: same here for `clearData()` and `initializeData()`

nemesisdesign (changes requested):
I do not see progress with this one, I think instead this has gotten worse.
The rrror messages are not clear. Please make sure the error messages printed are short and clear, see below for more suggestions on this.

### Error when app is not running

If the app is not running, I get this error:

```
    WebDriverError: Reached error page: about:neterror?e=connectionFailure&u=http%3A//0.0.0.0%3A8080/default/login&c=UTF-8&d=Firefox%20can%E2%80%99t%20establish%20a%20connection%20to%20the%20server%20at%200.0.0.0%3A8080.

      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:517:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:642:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:568:28)
      at Driver.execute (node_modules/selenium-webdriver/lib/webdriver.js:731:17)


  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "# Shell Plus Model Imports
    from allauth.account.models import EmailAddress, EmailConfirmation
    from allauth.socialaccount.models import SocialAccount, SocialApp, SocialToken
    from django.contrib.admin.models import LogEntry
    from openwisp_users.models import Group, Organization, OrganizationOwner, OrganizationUser, User
    from django.contrib.auth.models import Permission
    from django.contrib.contenttypes.models import ContentType
    from django.contrib.sessions.models import Session
    from django.contrib.sites.models import Site
    from openwisp_radius.models import Nas, OrganizationRadiusSettings, PhoneToken, RadiusAccounting, RadiusBatch, RadiusCheck, RadiusGroup, RadiusGroupCheck, RadiusGroupReply, RadiusPostAuth, RadiusReply, RadiusToken, RadiusUserGroup, RegisteredUser
    from rest_framework.authtoken.models import Token, TokenProxy
    # Shell Plus Django Imports
    from django.core.cache import cache
    from django.conf import settings
    from django.contrib.auth import get_user_model
    from django.db import transaction
    from django.db.models import Avg, Case, Count, F, Max, Min, Prefetch, Q, Sum, When
    from django.utils import timezone
    from django.urls import reverse
    from django.db.models import Exists, OuterRef, Subquery
    Python 3.8.5 (default, May 27 2021, 13:30:53) 
    Type 'copyright', 'credits' or 'license' for more information
    IPython 7.18.1 -- An enhanced Interactive Python. Type '?' for help.

    In [1]: 
    In [2]: 
    In [2]: 
    In [3]: 
    In [3]: Out[3]: <OrganizationUser:  (default)>

    In [4]: Do you really want to exit ([y]/n)? 
    ".

      18 |     // });
      19 |     await executeCommand("./browser-test/initialize_data.py", (message) => {
    > 20 |       console.log(message);
         |               ^
      21 |     });
      22 |     driver = await getDriver(Builder, new firefox.Options().headless());
      23 |   });

      at console.log (node_modules/@jest/console/build/BufferedConsole.js:197:10)
      at callback (browser-test/login.test.js:20:15)
      at browser-test/utils.js:31:12
```

### Build output on github actions

Even the passing output on github actions is messy, please take a look at the output of browser tests:
https://github.com/openwisp/openwisp-wifi-login-pages/pull/261/checks?check_run_id=2884204133

This is very wrong Sankalp, passing tests shouldn't return so much output.

browser-test/clear_data.py
```diff
@@ -2,25 +2,34 @@
...
+else:
+    process = subprocess.Popen(
+        command, shell=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    stdout, stderr = process.communicate()
+    errors.extend(re.findall('(\\n.*?Error.*?\\n)', stderr.decode('utf-8')))
+    sys.stdout.write(stdout.decode('utf-8'))
+sys.stderr.write('\n'.join(errors))
```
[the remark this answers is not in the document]
codesankalp: Environment variable is still used in django.setup()

nemesisdesign (changes requested):
Almost there, please resolve the issues highlighted in the comments below.

browser-test/initialize_data.py
```diff
@@ -48,13 +48,18 @@ def load_test_data():
...
 
 user = User.objects.create_user(
     username=test_user_email, password=test_user_password, email=test_user_email
 )
-org = Organization.objects.get(name=test_user_organization)
+try:
+    org = Organization.objects.get(slug=test_user_organization)
+except:
+    print(
+        f'The organization {test_user_organization} does not exist in the OpenWISP Radius environment specified ({OPENWISP_RADIUS_PATH}), please create it and repeat the tests.',
```
nemesisdesign, line 61: break this line over two lines please and we're ready to go

nemesisdesign (approved):
:+1:

Issue #236, opened 2021-05-18
[Feature] Add support for default configuration and reduce boilerplate


codesankalp:
Avoid having to repeat the whole configuration options: right now the configuration of each organization contains a lot of boilerplate. We shall introduce default configurations and ensure the application works also when the configuration file of a specific organization misses a piece of configuration.

When the ability to remove specific sections or fields is needed, right now we resorted to deleting the specific part of the configuration, but once we introduce this change we will have to ensure the configuration options that would have been removed can be set to null to obtain the same result

Issue #235, opened 2021-05-18
[feature] Implement get text like translations


codesankalp:
Implement get a text like translations: right now translations have to be defined in the configuration file of each organization, repeating the same text over and over, we should avoid this and store the translations in a central place;

However, being able to customize the text for each organization is a great feature and should still be possible if needed.

nemesisdesign:
I'd like to ask suggestions from @R9295 with whom I am working on some other project in which we used https://ttag.js.org/ to do this.

Aarnav, here we'll need to do the same thing, with the difference that texts can be overridden by configuration if needed.
If you have any suggestion or recommendation, let us know!

R9295:
Since there is an express backend, you could lazy load the files that ```ttag``` needs (```.po.json```) depending on the locale, which is ideal. I will think a bit more about dynamic po file generation based on the config.

nemesisdesign:
What if we define all strings as placeholders and then have po files for each language, including english?
We could then tweak the code to look for custom po files first, and if the customized versions are found the default po files are ignored.

For example, by default the app would look for:

1. `{languagecode}_custom.po`
2. `{languagecode}.po`

By default option 1 will not be provided, but users of this app can prepare a custom po file with different wording if needed.
That means that by default, option 2 is used.

PR #264, opened 2021-06-07
[feature] Password with eye icon

### LINKED ISSUE #263: [feature] Password with eye icon
On mobile devices dealing with the password fields can be annoying, we could create a simple component that handles this, or reuse a third party solution (provided it's lightweight, eg: https://github.com/psd8/react-password-toggle-icon).

We could then use this component in:

- login
- registration
- change password
- confirm reset password

codesankalp:
Closes #263


.gitignore
```diff
@@ -34,6 +34,7 @@ yarn-error.log*
 client/config.json
 !client/assets/
 client/assets/*
+!client/assets/**.svg
```
nemesisdesign, line 37: are you sure it wouldn't better to explicitly specify the eye icons? Otherwise don't we risk someone to commit their own SVG files into this repo?

client/index.css
```diff
@@ -265,3 +265,19 @@ h1 {
     font-size: 16px;
   }
 }
+.eye {
+  content: url("assets/eye.svg");
+  width: 30px;
```
nemesisdesign, line 270: the width can be made a bit smaller

client/index.css
```diff
@@ -265,3 +265,19 @@ h1 {
...
+  float: right;
+  margin-top: -35px;
+  margin-right: 5px;
```
nemesisdesign, line 273: we can increase a bit margin on the right to increase whitespace

client/index.css
```diff
@@ -265,3 +265,19 @@ h1 {
...
+  position: relative;
+}
+.eye-slash {
+  content: url("assets/eye-slash.svg");
+  width: 30px;
+  float: right;
+  margin-top: -35px;
+  margin-right: 5px;
+  position: relative;
```
nemesisdesign, line 282: same on this icon

client/utils/password-toggle.js
```diff
@@ -0,0 +1,69 @@
...
+            : "password",
+        );
+      },
+    );
+  }
+
+  render() {
+    const {inputRef, parentClassName} = this.props;
+    const showPasswordIcon = () => <i className="eye" />;
+    const hidePasswordIcon = () => <i className="eye-slash" />;
```
nemesisdesign, line 40: can you please add a `title` attribute so that if the user leaves the mouse on the icon a hint will appear?

Eg:

- reveal password
- hide password

The only problem is how to make this strings translatable, with the current system we'd have to define it in the YAML, maybe defining it just once and passing that value to the component is enough for now, but for sure soon we'll have to start working on the [translation issue](https://github.com/openwisp/openwisp-wifi-login-pages/issues/235).

.gitignore
nemesisdesign, line 37: are you sure it wouldn't better to explicitly specify the eye icons? Otherwise don't we risk someone to commit their own SVG files into this repo?

nemesisdesign (changes requested):
Great work @codesankalp!

The icon chosen is a bit intimidating though:

Screenshot from 2021-06-07 16-47-25

Can you find an icon which is more similar to the following?



Or anything less intimidating! :joy: 

See below for further suggestions.

PS: please ensure your commits follow the commit message guidelines for referencing and closing issues:
https://openwisp.io/docs/developer/contributing.html#commit-message-style-guidelines

codesankalp:
I will change the first commit message after the review/

internals/generators/organization/config.yml.hbs
```diff
@@ -518,6 +518,13 @@ client:
       additional_fields: []
       logout_by_session: {{logout_by_session_ID}}
 
+  password_eye_icon:
```
nemesisdesign, line 521: Just to make you understand why reworking the conf will be important I'm leaving this comment:
now that I tested this branch on a config that was created before this, the piece of code which calls this fails because it doesn't find it.

There's not action to do here, it's good enough for now, we'll improve these aspects in https://github.com/openwisp/openwisp-wifi-login-pages/issues/236 and https://github.com/openwisp/openwisp-wifi-login-pages/issues/235.

nemesisdesign (changes requested):
Looks great!

One additional minor request for improvement: can you make the eye icon cursor clickable using CSS `cusor: pointer` please?

Also fix conflicts.

Then we'll be ready to merge this as well :+1:.

nemesisdesign (approved):
Thanks! :+1:

PR #268, opened 2021-06-12
[feature] Set page title for each page #155

### LINKED ISSUE #155: [wifi-login-pages] Set page title
For each page of the app we should have a title, eg: `{page_title} - {organization_name}`.

We should have a `title` attribute in the org-config of each component, translatable, eg:
```yaml
registration_form:
    title:
        en: "Sign up"
```

We should add a test for each component to ensure the tile is changed as expected.

Also the main title should be allowed to be multilingual, now we have:

```yaml
client:
  # <title> value
  title: "WiFi Login"
```

we shall switch to:
```yaml
client:
  # <title> value
  title:
    en: "WiFi Login"
    it: "Accesso WiFi"
    # other languages...
```

codesankalp:
Added translatable support for page title in each page.

Closes #155


nemesisdesign (commented):

client/components/registration/registration.js
```diff
@@ -1034,4 +1039,7 @@ Registration.propTypes = {
     content: PropTypes.object,
   }).isRequired,
   authenticate: PropTypes.func.isRequired,
+  setUserData: PropTypes.func.isRequired,
+  userData: PropTypes.object.isRequired,
```
nemesisdesign, line 1043: pay attention, I removed `userData` and `setUserData` from this component, I centralized all the operations in the status component to simplify and avoid duplication, therefore these lines should not be added back.

client/components/registration/registration.test.js
```diff
@@ -23,11 +23,15 @@ const createTestProps = function (props, configName = "default") {
...
     orgSlug: configName,
+    orgName: config.name,
     settings: config.settings,
     registration: config.components.registration_form,
     privacyPolicy: config.privacy_policy,
     termsAndConditions: config.terms_and_conditions,
     authenticate: jest.fn(),
+    userData: {},
+    setUserData: jest.fn(),
+    setTitle: jest.fn(),
```
nemesisdesign, line 34: same here

client/components/registration/index.js
```diff
@@ -11,12 +11,15 @@ const mapStateToProps = (state) => {
...
     orgSlug: conf.slug,
+    orgName: conf.name,
     language: state.language,
   };
 };
 const mapDispatchToProps = (dispatch) => {
   return {
     authenticate: authenticate(dispatch),
+    setUserData: setUserData(dispatch),
+    setTitle: setTitle(dispatch),
```
nemesisdesign, line 22: and here

nemesisdesign (changes requested):
This is also almost ready.
See my notes below.

Please **note** my minor changes here: https://github.com/openwisp/openwisp-wifi-login-pages/pull/268/commits/bc731370721657875dd50e5e3aa813311c69fd41

client/components/login/login.js
```diff
@@ -48,7 +48,9 @@ export default class Login extends React.Component {
   componentDidMount() {
     const username = getParameterByName("username");
     const token = getParameterByName("token");
-    const {loginForm} = this.props;
+    const {loginForm, setTitle, orgName, language} = this.props;
+    setTitle(`${getText(loginForm.title, language)} - ${orgName}`);
```
nemesisdesign, line 52: Why don't we change the `setTitle` function to accept different parameters?

eg: 

```
setTitle(component, language, orgName)
```

This allows to avoid compiling the title string in many different places so if we want to change its general structure we can do it from one place.

nemesisdesign (approved):

PR #269, opened 2021-06-14
[feature] Allow specifying if link should be shown to verified users #107

### LINKED ISSUE #107: [wifi-login-pages] Allow specifying if link should be shown to verified users
To be done after #100.
Follow up of #93.

Thanks to #93 we can now specify if a link should be shown to:

- anyone
- only authenticated users
- only unauthenticated users

However, we need to show specific links only in some cases.
For example, the "change mobile phone number" makes sense only when the registration method the user chose is `mobile_phone`.

Other links like change password make sense only when the user is already verified.
 
We can add a few additional properties to specify it the link should be shown only to specific type of users, eg:

```
# any authenticated user
authenticated: true

# only authenticated users which registered with mobile phone verification
authenticated: true
method: 'mobile_phone'

# only verified and authenticated users
authenticated: true
verified: true

# only unverified authenticated users
authenticated: true
verified: false
```

The new property makes sense only when `authenticated: true` is supplied and should be ignored otherwise.

codesankalp:
Passed userData in shouldLinkBeShown to handle links for verified users.

Closes #107

### DIFF
--- a/client/utils/should-link-be-shown.js
+++ b/client/utils/should-link-be-shown.js
@@ -1,4 +1,21 @@
-const shouldLinkBeShown = (link, isAuthenticated) => {
+/* eslint-disable camelcase */
+const shouldLinkBeShown = (link, isAuthenticated, userData) => {
+  const {is_verified} = userData;
+  if (
+    link.authenticated === isAuthenticated &&
+    isAuthenticated === true &&
+    link.verified !== undefined &&
+    is_verified !== undefined
+  ) {
+    return link.verified === is_verified;
+  }
+  if (
+    link.authenticated === isAuthenticated &&
+    isAuthenticated === true &&
+    link.method !== undefined
+  ) {
+    return link.method === "mobile_phone";
+  }
   return (
     link.authenticated === undefined || link.authenticated === isAuthenticated
   );
--- a/client/utils/should-link-be-shown.js
+++ b/client/utils/should-link-be-shown.js
@@ -14,7 +14,7 @@ const shouldLinkBeShown = (link, isAuthenticated, userData) => {
     isAuthenticated === true &&
     link.method !== undefined
   ) {
-    return link.method === "mobile_phone";
+    return link.method === userData.method;
   }
   return (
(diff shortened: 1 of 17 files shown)


nemesisdesign (changes requested):

codesankalp:
I have tested it for default and mobile verification.

client/utils/should-link-be-shown.js
```diff
@@ -1,4 +1,21 @@
...
+    is_verified !== undefined
+  ) {
+    return link.verified === is_verified;
+  }
+  if (
+    link.authenticated === isAuthenticated &&
+    isAuthenticated === true &&
+    link.method !== undefined
+  ) {
+    return link.method === "mobile_phone";
```
nemesisdesign, line 17: This is not working according to my test.
And indeed it looks quite wrong doesn't it? Why do you test if the link method is "mobile_phone"?
I think it should be: `return link.method === userData.method;`, this fixes the behavior in my manual testing, but please change your tests to catch this mistake and let me know which test is the failing one once it's done, so I will double check it.

nemesisdesign (changes requested):
Almost.. found one issue.

client/utils/utils.test.js
```diff
@@ -143,14 +143,22 @@ describe("shouldLinkBeShown tests", () => {
...
     expect(shouldLinkBeShown(link, isAuthenticated, userData)).toBe(false);
   });
+  it("should return true if userData.method and link.method is same", () => {
+    const {link, isAuthenticated, userData} = createArgs(
+      {authenticated: true, method: "payment"},
+      true,
+      {method: "payment"},
+    );
+    expect(shouldLinkBeShown(link, isAuthenticated, userData)).toBe(true);
+  });
```
codesankalp: This is the failing test if we remove the changes made in commit https://github.com/openwisp/openwisp-wifi-login-pages/pull/269/commits/5d5161a3434aa620ece014144efdc9da4ded46b5

nemesisdesign (approved):

PR #271, opened 2021-06-14
[fix] Changed is_verified to default to null #270

### LINKED ISSUE #270: [bug] is_verified defaults to true
@codesankalp try the following patch:

```diff
diff --git a/client/components/status/status.js b/client/components/status/status.js
index 5e7131c..6c221e5 100644
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -88,6 +88,9 @@ export default class Status extends React.Component {
         logout,
       );
       setLoading(false);
+
+      console.log(userData);
+
       if (isValid) {
         const {justAuthenticated} = userData;
         ({userData} = this.props);
```

Now log in, first time the user data is logged in the console correctly.

Now reload the page, this time the user data is not logged and the default values are used instead.

This is giving me pain while working on https://github.com/openwisp/openwisp-wifi-login-pages/pull/170.

**Question 1**: why is that `console.log` returning the default data instead of the data we received from the server?
**Question 2**: how to we access the data which was returned from the server instead of the default data?

**Question 3**: can you explain why you default `is_verified` to `true` here?

https://github.com/openwisp/openwisp-wifi-login-pages/blob/0caf7ddeb0eb126a1eb35104f7638ec8bd8f5aea/client/reducers/organization.js#L18-L27

**Question 4**: why is this information duplicated 2 more times?

https://github.com/openwisp/openwisp-wifi-login-pages/blob/0caf7ddeb0eb126a1eb35104f7638ec8bd8f5aea/client/actions/set-organization.js#L23-L27

https://github.com/openwisp/openwisp-wifi-login-pages/blob/0caf7ddeb0eb126a1eb35104f7638ec8bd8f5aea/client/actions/set-organization.js#L43-L45

We should define this information only once and I believe `is_verified` should default to `null`.

nemesisdesign:
Fixes #270

### DIFF
--- a/client/actions/actions.test.js
+++ b/client/actions/actions.test.js
@@ -48,7 +48,7 @@ describe("actions testing", () => {
     const orgConfig2 = merge(testOrgConfig[0], testOrgConfig[1]);
     const userData = {
       is_active: true,
-      is_verified: true,
+      is_verified: null,
       justAuthenticated: true,
     };
     orgConfig.userData = userData;
--- a/client/actions/set-organization.js
+++ b/client/actions/set-organization.js
@@ -11,6 +11,7 @@ import authenticate from "../utils/authenticate";
 import customMerge from "../utils/custom-merge";
 import getConfig from "../utils/get-config";
 import logout from "./logout";
+import {initialState} from "../reducers/organization";
 
 const setOrganization = (slug, cookies) => {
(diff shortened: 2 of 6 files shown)


codesankalp (approved):
Tested locally,
Working fine on mobile verification needed organization and default organization.

Issue #272, opened 2021-06-16
[change] Move redirect logic from OrganizationWrapper to each component


nemesisdesign:
I suspected that it was wrong to define the redirection logic in `OrganizationWrapper` but now I think I had the confirmation.
Basically the organization wrapper files decides if the user should be redirected to verification or not and so on.
But that component is re-rendered everytime `setLoading()` is called.

I found out while working on #170 because I need to redirect the user to the credit card verification which is an external link, and in the automated tests I found out the redirection was called twice.

In reality I found out that multiple HTTP requests were being made to the payment URL, which would have very bad consequences in production because it would initiate multiple payments to the payment gateway.

I think we will have to change this part and move most of that logic in each individual component, this also makes it easier to test it.

Making OrganizationWrapper leaner should also have a beneficial effect on rendering performance because less logic will be executed each time `setLoading()` is called.

@codesankalp we can work on this if there's still time as last thing.

codesankalp:
I agree with this, redirection must be handled from the component so that routes defined in the organization wrapper will be more cleaner. Because currently, we check for some conditions and then redirect.
I also face multiple redirections when doing mobile verification, which can also be solved using this.
- I will also implement proper routings such as AuthRoute and simple Route which will make it better.

I will start working on this.

vikram-2101:
Hi @nemesifier ,

I've been studying this issue and the OrganizationWrapper source. I can see the redirect logic in organization-wrapper/index.js triggers on every setLoading() call, which is the root of the multiple-redirect bug you described.

I'd like to propose and implement moving these redirects:

- Verification redirect → into mobile-phone-verification and password-confirm components
- Payment/credit card redirect → into payment-process and registration components
- General auth redirect → into status component's componentDidMount

Each moved redirect would also get its own test. Is this the right direction? If so, I can open a draft PR showing the move for one component first so you can validate the approach before I do all of them.

PR #276, opened 2021-06-16
[tests] Added basic tests for status route

nemesisdesign:
I added these tests yesterday because I needed to add a test for ensuring credit card verification is initiated, and that's when I come to the conclusions explained in https://github.com/openwisp/openwisp-wifi-login-pages/issues/272, so I ended up moving the logic which initiates the credit card verification in the status component and moved the test there as well (I think we should concentrate the logic which initiates identity verification in the status component and have all other modules simply redirect to status to initiate verification, it's a lot simpler).

However, I did quite some work to figure out how to test the behavior of routes so I don't want to throw this away because we have basically no tests for routes and we should have more.

### DIFF
--- a/client/components/organization-wrapper/organization-wrapper.test.js
+++ b/client/components/organization-wrapper/organization-wrapper.test.js
@@ -1,8 +1,27 @@
 /* eslint-disable camelcase */
-import {shallow} from "enzyme";
-import React from "react";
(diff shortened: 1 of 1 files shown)


codesankalp (approved):
Tested and looks good, test coverage also got increased :smile:

PR #277, opened 2021-06-18
[Fix] Added wait after captive portal login form submission #265

### LINKED ISSUE #265: [bug] Current active session not shown after login
Depends on #259 for testing.

Once the user logs in and captive portal login is performed, the new open session should be shown in the UI but it's not, because sessions are loaded before the captive portal login.

Ideally we should load sessions only after the captive portal login has finished loading, we could even wait 1 extra second just to be sure.
Alternatively, we can reload sessions again, but we're already making a lot of requests and I'd avoid that.

codesankalp:
Fixes #265

### DIFF
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -136,6 +136,8 @@ export default class Status extends React.Component {
             this.loginFormRef.current.submit();
             setUserData({...userData, justAuthenticated: false});
           }
+          // waiting for 1 second after submitting captive portal login form
+          await new Promise((res) => setTimeout(res, 1000));
           await this.getUserActiveRadiusSessions();
           await this.getUserPassedRadiusSessions();
           const intervalId = setInterval(() => {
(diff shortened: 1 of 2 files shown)


client/components/status/status.js
```diff
@@ -136,6 +136,8 @@ export default class Status extends React.Component {
             this.loginFormRef.current.submit();
             setUserData({...userData, justAuthenticated: false});
           }
+          // waiting for 1 second after submitting captive portal login form
+          await new Promise((res) => setTimeout(res, 1000));
```
nemesisdesign, line 140: we do not know whether it will take 1 second, less or more.
We have to find a better way to do this. I think we have to add an onload eventl listener to the iframe.
I think we have to do this both for login and logout, with a timeout which defaults to 5 seconds and if reached gives a specific error message, something like (captive portal is not reachable).

nemesisdesign (changes requested):

client/components/status/status.js
codesankalp: Ok, I will add this change.

nemesisdesign:
Closing as I'm taking care of this.

PR #282, opened 2021-06-25
[tests] Fixed browser tests #280

### LINKED ISSUE #280: [CI] Driver is undefined in browser test
Sometimes jest `beforeAll` hook gave timeout in driver initialization.
We want to be specific while running tests.
`beforeAll` hook should wait till the driver is initialized and then the test must run.

The issue is mentioned here https://github.com/facebook/jest/issues/9527

We can use jest-circus as a test runner to solve this issue.

image

nemesisdesign:
Closes #280

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -34,6 +34,16 @@
     ],
     "react/jsx-props-no-spreading": 0,
     "react/forbid-prop-types": 0,
-    "react/sort-comp": [0]
+    "react/sort-comp": [0],
+    "import/no-extraneous-dependencies": [
+      "error",
+      {
+        "devDependencies": [
+          "**/*.test.js",
+          "**/jest.config.js",
+          "browser-test/utils.js"
+        ]
+      }
+    ]
   }
 }
--- a/browser-test/utils.js
+++ b/browser-test/utils.js
@@ -1,34 +1,31 @@
-import {spawn} from "child_process";
+import {Builder} from "selenium-webdriver";
+import {spawnSync} from "child_process";
 import testData from "./testData.json";
+
+const firefox = require("selenium-webdriver/firefox");
 
 const waitTime = 5000;
 const orgSlug = "default";
 
-export const executeCommand = async (command, argv) => {
-  const executor = await spawn(command, [argv]);
-  executor.stdout.on("data", (data) => {
-    process.stdout.write(`${data}\n`);
-  });
+export const executeCommand = (command, argv) => {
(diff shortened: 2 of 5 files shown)


browser-test/utils.js
```diff
@@ -1,34 +1,31 @@
...
+
+  if (result.stdout) {
+    process.stdout.write(`${result.stdout}\n`);
+  }
+  if (result.stderr) {
+    process.stderr.write(`${result.stderr}\n`);
+  }
+  if (result.status !== 0) {
+    process.exit(result.status);
+  }
```
nemesisdesign, line 21: @codesankalp unfortunately I am no expert when it comes to nodeJS.
While testing I have noticed that the scripts keep running after the tests have finished and jest complains about it.
After debugging I found out that we could have used `spawnSync` which is a lot simpler and more suited for this kind of operation.

I also have seen there's `execSync` but would have probably meant to change a lot more the code so I didn't go for it.

browser-test/utils.js
```diff
@@ -56,3 +53,5 @@ export const urls = {
   login: `http://0.0.0.0:8080/${orgSlug}/login`,
   status: `http://0.0.0.0:8080/${orgSlug}/status`,
 };
+
+jest.setTimeout(10000);
```
nemesisdesign, line 59: I increased the global jest time out here.

browser-test/utils.js
```diff
@@ -1,34 +1,31 @@
...
 
-export const getDriver = async (Builder, options) => {
-  const driver = await new Builder()
+export const getDriver = async () => {
+  return new Builder()
     .forBrowser("firefox")
-    .setFirefoxOptions(options)
+    .setFirefoxOptions(new firefox.Options().headless())
     .build();
-  return driver;
```
nemesisdesign, line 31: @codesankalp I simplified this. No need to repeat the headless option and `Builder` in each test file.

.eslintrc.json
```diff
@@ -34,6 +34,16 @@
     ],
     "react/jsx-props-no-spreading": 0,
     "react/forbid-prop-types": 0,
-    "react/sort-comp": [0]
+    "react/sort-comp": [0],
+    "import/no-extraneous-dependencies": [
```
nemesisdesign, line 38: this is needed after moving the imports of selenium in `utils.js`

browser-test/utils.js
codesankalp: Yes, it needs to be increased.

PR #284, opened 2021-06-26
[tests] Further improvements to browser tests

nemesisdesign:
@codesankalp let's avoid to leave unused code ok?


codesankalp:
> @codesankalp let's avoid leaving unused code ok?

I thought of removing it in the end but forgot to remove the unused code at that time. Sorry for this.
I have changed the code to use CSS as mentioned [here](https://github.com/openwisp/openwisp-wifi-login-pages/pull/286#discussion_r659255655).

PR #286, opened 2021-06-27
[change/fix] Wait until captive portal login response #265

### LINKED ISSUE #265: [bug] Current active session not shown after login

nemesisdesign:
Wait until captive portal login response to perform actions like:

- clearing the loading overlay
- redirecting the user to payment flow
- loading user sessions

Fixes #265

### DIFF
--- a/browser-test/login.test.js
+++ b/browser-test/login.test.js
@@ -44,5 +44,11 @@ describe("Selenium tests for <Login />", () => {
     await driver.wait(until.elementIsVisible(successToastDiv));
     await driver.wait(until.urlContains("status"), 5000);
     expect(await successToastDiv.getText()).toEqual("Login successful");
+
+    const activeSessionTr = await getElementByCss(
+      driver,
+      "table tr.active-session",
+    );
+    await driver.wait(until.elementIsVisible(activeSessionTr));
   });
 });
--- a/browser-test/clear_data.py
+++ b/browser-test/clear_data.py
@@ -4,6 +4,7 @@ import os
 import sys
 
 import django
+import swapper
 
 
 def load_test_data():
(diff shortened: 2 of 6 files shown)


browser-test/login.test.js
```diff
@@ -47,5 +48,11 @@ describe("Selenium tests for <Login />", () => {
     await driver.wait(until.elementIsVisible(successToastDiv));
     await driver.wait(until.urlContains("status"), 5000);
     expect(await successToastDiv.getText()).toEqual("Login successful");
+
+    const activeSessionTr = await getElementByCss(
+      driver,
+      "table tr.active-session",
+    );
+    await driver.wait(until.elementIsVisible(activeSessionTr));
```
nemesisdesign, line 56: unfortunately this doesn't fail for me without the fix, I was not able to replicate this issue in automated tests, only with manual testing :(

browser-test/utils.js
```diff
@@ -34,6 +34,12 @@ export const getElementByXPath = async (driver, xpath) => {
   return el;
 };
 
+export const getElementByCss = async (driver, css) => {
```
nemesisdesign, line 37: @codesankalp let's use css instead of xpath which is more universally known

codesankalp: Added it in this https://github.com/openwisp/openwisp-wifi-login-pages/pull/284

browser-test/login.test.js
codesankalp: Yes, I also tried but sessions are showing in the browser test.

codesankalp (approved):

PR #287, opened 2021-06-28
[feature] Implemented get text like translations #235

### LINKED ISSUE #235: [feature] Implement get text like translations

### LINKED ISSUE #304: [bug] Ensure language preferences are respected
The current app has two flaws in regard to translations:

- **it does not take in consideration the preferred language of the browser.**
  For example, if the app has 2 languages: English and Indian; while the browser of the user has set Indian as preferred language, then when the app is opened, it should load Indian automatically, without needing the user to select it.

- **if the user changes languages, the choice is forgotten at page reload**.
  Continuing the example above, say that for some reason the user prefers to see the English translation instead of the Indian one, once the user clicks to switch to English, the choice should be respected in subsequent visits / reloads, maybe we can simply use localStorage for this?

@codesankalp can you address these issues once #287 is sorted out please?

### LINKED ISSUE #308: [change] Pass current language to Intl.DateTimeFormat
To be done after #235 and #304.

There's a few lines of code which pass a language code to format dates:

https://github.com/openwisp/openwisp-wifi-login-pages/blob/0de45a21ff10cb39967e18578e9e6106dffe0a3f/client/components/status/status.js#L457-L468

https://github.com/openwisp/openwisp-wifi-login-pages/blob/0de45a21ff10cb39967e18578e9e6106dffe0a3f/client/components/status/status.js#L512-L531

However, this langauge code is not dynamic, it's statically configured. We should be able to pass the current language used in the app (depending on user preference) to this function.

codesankalp:
Used ttag js to implement get text like translations.

Closes #235
Fixes #304 
Closes #308


nemesisdesign (changes requested):
Fix the conflict please.

.gitignore
```diff
@@ -38,6 +38,9 @@ client/assets/*
 !client/assets/eye-slash.svg
 !client/assets/default/
 !client/assets/fonts/
+client/translations/
+i18n/*
+!i18n/en.po
```
nemesisdesign, line 43: shouldn't this be `!i18n/*.po`?
Otherwise we would include only english translations.

README.md
```diff
@@ -270,41 +270,33 @@ Let us consider the following configuration for the header, footer and contact c
...
-          en: "change password"
+      - text: "change password"
         url: "/change-password"
         authenticated: true
         # if organization supports any verification method
         verified: true
       # if organization supports mobile verification
-      - text:
-          en: "change phone number"
+      - text: "change phone number"
```
nemesisdesign, line 284: shouldn't we be able to remove this text from here?

README.md
```diff
@@ -270,41 +270,33 @@ Let us consider the following configuration for the header, footer and contact c
...
         authenticated: true
   contact_page:
     social_links:
-      - text:
-          en: "support"
+      - text: "support"
         url: "/support"
-      - text:
-          en: "twitter"
+      - text: "twitter"
```
nemesisdesign, line 299: same here

README.md
```diff
@@ -403,6 +389,37 @@ enable this feature, just create a new organization with the
 
 `Are you using OpenWISP Subscriptions to provide paid subscriptions for WiFi plans or identity verification via credit/debit card?`
 
+#### Translation
```
nemesisdesign, line 392: Translation > Translations?

README.md
```diff
@@ -403,6 +389,37 @@ enable this feature, just create a new organization with the
...
+
+1. All the translation is loaded at runtime from the JSON files that were made
```
nemesisdesign, line 394: translations are
were compiled instead of were made

README.md
```diff
@@ -403,6 +389,37 @@ enable this feature, just create a new organization with the
...
+   `{language_code}.po`.
+
+6. Translations inside `i18n/` directory can be updated by using this command:
+   `yarn update-translation i18n/{file_name}.po client/`
+
+   This will update all the translations by replacing the placeholder defined in the `client/`.
+
+7. If there is more than one language in `i18n/` directory then update the organization configuration file by adding the support for that language like this:
+
+   ```
```
nemesisdesign, line 415: let's flag as yaml to get the colored syntax

package.json
```diff
@@ -107,7 +110,9 @@
     "test": "jest --testPathIgnorePatterns=/browser-test/",
     "browser-test": "jest browser-test --runInBand",
     "test-inspect": "node inspect ./node_modules/.bin/jest --runInBand",
-    "stats": "STATS=true npx webpack-cli --profile --json --config config/webpack.config.js > compilation-stats.json && webpack-bundle-analyzer dist/stats.json"
+    "stats": "STATS=true npx webpack-cli --profile --json --config config/webpack.config.js > compilation-stats.json && webpack-bundle-analyzer dist/stats.json",
```
nemesisdesign, line 113: why npx?

nemesisdesign (commented):
Terms & conditions and Privacy Policy: there's a problem here.
These will surely vary in each installation, so there's no point in adding these to the list of translatable texts.
Can you find a way to keep these as configuration please?

nemesisdesign (changes requested):
Another potential issue with the way language files are built and included:
I see that the language strings are included in the main chunk, but what happens if there are multiple languages, say 5 languages? Would all 5 langauges be included in the main chunk?

We should make sure each language is stored in its own file and then should be loaded only if the user is opening the app with that language.

i18n/en.po
```diff
@@ -0,0 +1,660 @@
...
+msgid "CONTACT_SOCIAL_LINK ${ link.alt }"
+msgstr "${ link.alt }"
+
+#: client/components/contact-box/contact.js:20
+msgid "CONTACT_EMAIL_LABEL"
+msgstr "E-mail"
+
+#: client/components/contact-box/contact.js:22
+msgid "CONTACT_EMAIL"
+msgstr "support@openwisp.co"
```
nemesisdesign, line 628: :man_facepalming:  :facepalm:  :face_with_head_bandage:  :face_with_thermometer: 
these kind of things should always be configuration... Applies to any other organization specifc detail.

i18n/en.po
```diff
@@ -0,0 +1,660 @@
...
+msgid "404_PAGE_MESSAGE"
+msgstr "Sorry, we couldn't find the page you were looking for."
+
+#: client/components/404/404.js:38
+msgid "HOME_PAGE_LINK_TEXT"
+msgstr "Go back to homepage"
+
+#: client/components/status/status.js:602
+msgid "DATE_LANGUAGE_LOCALE"
+msgstr "en-US"
```
nemesisdesign, line 660: how did this end up here? Does not look like a translatable string, it's just the language code, can you point out at what part of the code is generating this please?

internals/generators/organization/config.yml.hbs
```diff
@@ -67,297 +64,133 @@ client:
         alternate_text: "{{name}}"
 {{/if}}
       links:
-        - text:
-            en: "Sign In" #link text in english
+        - text: "Sign In" #link text in english
```
nemesisdesign, line 67: mmm...

What if a user wants to add a new link and translate it in multiple languages?
Can they do it somehow?

README.md
```diff
@@ -403,6 +389,37 @@ enable this feature, just create a new organization with the
...
+   languages:
+     - text: "english"
+       slug: "en"
+     - text: "Spanish"
+       slug: "es"
+   ```
+
```
nemesisdesign, line 422: What about the possibility of changing/customizing translations?

nemesisdesign (changes requested):

i18n/en.po
codesankalp: I also don't want to define it here, but In configuration, I saw email is also defined for multiple languages i.e. old getText is used in the code for email value also.

codesankalp: I am making email for single language only.

internals/generators/organization/config.yml.hbs
codesankalp: Used previous getText

i18n/en.po
codesankalp: In status page for account start and stop time. Basically it translates the time to be in the format of that locale

README.md
codesankalp: updated readme for this.

client/components/organization-wrapper/organization-wrapper.test.js
```diff
@@ -214,7 +212,7 @@ describe("Test <OrganizationWrapper /> routes", () => {
 
   it("should display status if authenticated", () => {
     wrapper = mountComponent(props, ["/default/status"]);
-    expect(mapRoutes(wrapper)["/default/status"]).toBe(Suspense);
+    expect(mapRoutes(wrapper)["/default/status"]).toBe(undefined);
```
codesankalp, line 215:
```suggestion
    expect(mapRoutes(wrapper)["/default/status"]).toBe(Suspense);
```
This is not giving Suspense when I tried. Do you know how to fix this?
Also, the component.find(Routes) returns empty but routes are present in that.
This is the only test where I faced problems in fixing.

nemesisdesign (changes requested):
Good progress @codesankalp :clap: :+1: !

There's a few minor conflicts to resolve due to the changes done in https://github.com/openwisp/openwisp-wifi-login-pages/pull/309, please take a look at the diff in the linked PR before merging the conflict.

See belo some issues I found during this round of testing.
This review is not exhaustive, I will be testing this work more in the next days while I work on the Italian translations.

### Be forgiving of not yet existing language translations

I had some translations defined in one of my testing confs, now the page does not load and this exception is logged in the JS console:

```
.*\.json$ namespace object:10 Uncaught (in promise) Error: Cannot find module './it.json'
    at .*\.json$ namespace object:10
```

Is it possible to change this part so that if the translation file is not found, the default language is used and a message is logged to console (eg: `console.error`)?

I have another question, what happens if a new string is added to the code of the application, but the translation is not yet available for every language? Will the code crash or will it display something anyway (and a message is logged to the console)? 

### Changing lang for an org changes it also for another org

Try having two orgs defined:

- default (leave default conf)
- second (add a lanaguage)

Now open the second org in a new tab and change the language to the other language you added, then go back to the tab with the default org and reload the page, what do you see?

I see this:

Screenshot from 2021-07-12 21-59-44

### Lang preferences

I see that you have addressed the second point of this issue: https://github.com/openwisp/openwisp-wifi-login-pages/issues/304, but you have not addressed the first, is that on purpose? We can do work on that after PR is complete if you prefer, I am just double checking with you.

client/components/logout/__snapshots__/logout.test.js.snap
```diff
@@ -23,7 +23,39 @@ exports[`<Logout /> rendering should render correctly 1`] = `
...
+    id="logout"
+  >
+    <div
+      className="inner"
+    >
+      <div
+        className="main-column w-100"
+      >
+        <h2>
+          LOGOUT_CONTENT
```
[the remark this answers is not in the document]
codesankalp: I added a test for rendering with placeholder tags, so that in the future if placeholder tags are changed then the snapshot must fail, and updating `po` files are required. Same reason for other occurences also.

README.md
```diff
@@ -420,6 +414,78 @@ enable this feature, just create a new organization with the
...
+1. Translations are loaded at runtime from the JSON files that were compiled during the build process.
+2. By default, the build process uses the `{language_code}.po` files to generate
+   the translations.
+3. `{language_code}.custom.po` file can be used to override the `{language_code}.po`
+   file. The custom files need not to be duplicate of the default file i.e. translation
+   can be defined as the custom strings.
+4. Translation can also be generated for single organisation if `{orgSlug}_{language_code}.custom.po` is present in `i18n` Directory.
+5. Translation can be created by running the command:
+   `yarn create-translation {language_code} i18n/{file_name}.po`
```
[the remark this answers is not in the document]
codesankalp: for custom po files, all the msg-id is not required, so  I thought to add a script just to create a translation file with content headers.

codesankalp:
> I have another question, what happens if a new string is added to the code of the application, but the translation is not yet available for every language? Will the code crash or will it display something anyway (and a message is logged to the console)?

It will display the translation placeholder on the webpage.

client/components/status/status.js
```diff
@@ -650,6 +629,41 @@ export default class Status extends React.Component {
...
+        start_time: t`ACCT_START_TIME`,
+        stop_time: t`ACCT_STOP_TIME`,
+        duration: t`ACCT_DURATION`,
+        download: t`ACCT_DOWNLOAD`,
+        upload: t`ACCT_UPLOAD`,
+        device_address: t`ACCT_DEVICE_ADDRESS`,
+      },
+      settings: {
+        active_session: t`ACCT_ACTIVE`,
+        date_language_locale: t`DATE_LANGUAGE_LOCALE`,
```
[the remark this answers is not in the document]
codesankalp: I will solve this in this.

README.md
```diff
@@ -420,6 +414,78 @@ enable this feature, just create a new organization with the
...
+
+   Here `file_name` can be `{orgSlug}_{language_code}.custom.po`, `{language_code}.custom.po\` or
+   `{language_code}.po`.
+
+6. Translations inside `i18n/` directory can be updated by using this command:
+   `yarn update-translation i18n/{file_name}.po client/`
```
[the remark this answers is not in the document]
codesankalp: Resolved

README.md
```diff
@@ -420,6 +414,78 @@ enable this feature, just create a new organization with the
...
+
+7. If there is more than one language in `i18n/` directory then update the organization configuration file by adding the support for that language like this:
+
+   ```yaml
+   languages:
+     - text: "english"
+       slug: "en"
+     - text: "Spanish"
+       slug: "es"
+   ```
```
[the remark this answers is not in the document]
codesankalp: Yes, it's still supported.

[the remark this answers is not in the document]
codesankalp: Like it will use the default if browser returns none in `navigator.language`

package.json
```diff
@@ -18,6 +18,7 @@
     "history": "^4.10.1",
     "js-yaml": "^3.14.0",
     "jsdom": "^16.4.0",
+    "lodash": "^4.17.21",
```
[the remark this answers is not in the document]
codesankalp: It's used in the build process only.

codesankalp:
> I see that you have addressed the second point of this issue: #304, but you have not addressed the first, is that on purpose? We can do work on that after PR is complete if you prefer, I am just double checking with you.

I implemented first part also, by using navigator. See the code in `load-translation.js`

```
if (useBrowserLang) lang = navigator.language || navigator.userLanguage;
```

package.json
codesankalp: I removed it.

README.md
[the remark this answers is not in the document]
nemesisdesign: @codesankalp it sounds like we should make two separate commands, one for adding general translations and one for adding custom translations, then document these differently and explain the difference, how does this sound?

client/utils/load-translation.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+};
+
+const loadTranslationUtil = async (
+  language,
+  orgSlug,
+  useBrowserLang = false,
+) => {
+  let lang = language;
+  if (useBrowserLang) lang = navigator.language || navigator.userLanguage;
```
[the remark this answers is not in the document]
codesankalp: Yes.

client/components/404/404.js
```diff
@@ -4,39 +4,36 @@ import "./index.css";
...
   render() {
-    const {orgSlug, language, page} = this.props;
+    const {orgSlug, page} = this.props;
     return (
       <div className="container content" id="not-foud-404">
         <div className="inner">
           <div className="main-column">
             <div className="row owisp-404-row-1">
-              {page && page.heading ? getText(page.heading, language) : "Oops!"}
+              {page && page.heading ? t`404_HEADING` : "Oops!"}
```
nemesisdesign: I think we can remove these default values at the end right? Let's try to shorten the code a bit where the old code is not necessary anymore, keep in mind we have to start thinking in reducing the size of the build.

nemesisdesign (changes requested):
Great progress @codesankalp.

I have done some improvements and pushed them to this branch, I also have the Italian translation file ready but I will send it in a separate PR when we are done with this.

While testing, I found quite a few non translated words in this file:
https://github.com/openwisp/openwisp-wifi-login-pages/blob/b6ef54a226fb2eb348e682041d73d6f7524f7e98/client/constants/index.js

See my comments below.

client/utils/load-translation.js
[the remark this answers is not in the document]
nemesisdesign: @codesankalp I verified that having "en-US" as preferred language breaks this code

nemesisdesign (changes requested):
BTW, having custom translations breaks the test "load translation tests › should load". Is there anything you can do to prevent that?
It will be annoying for people like me who have to maintain different services and may want to keep the custom translations in the same directory without committing them.

client/utils/load-translation.js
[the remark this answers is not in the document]
codesankalp: Resolved

client/components/404/404.js
codesankalp: I think No, because these default values are used on those URL where organization-slug is missing like `http://0.0.0.0:8080/error/` as setOrganization is not called. Also 404 page of organization and without organization is different.

README.md
```diff
@@ -420,6 +414,100 @@ enable this feature, just create a new organization with the
...
+`{language_code}.po`.
+
+See [update-translations](#update-translations) for extracting translation placeholders
+from the code.
+
+#### Update translations
+
+For extraction or updating translations in the `.po` file, Use the command:
+
+`POFILE=<path-to-po-file> yarn translations-update`
```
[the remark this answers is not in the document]
codesankalp: Because the first argument needs to be the path to po file and the second argument needs to be src files.

https://github.com/ttag-org/ttag-cli/blob/76b340e87b550193da03b9a3f88dbe88b648fc93/src/index.ts#L226-L236

internals/generators/organization/config.yml.hbs
```diff
@@ -96,260 +93,107 @@ client:
...
     registration_form:
-      title:
-        en: "Sign up"
       input_fields:
         phone_number:
-          placeholder:
-            en: "mobile phone number"
-          label:
-            en: "mobile phone number"
+          type: "tel"
```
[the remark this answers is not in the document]
codesankalp: Removed

codesankalp:
> BTW, having custom translations breaks the test "load translation tests › should load". Is there anything you can do to prevent that?
> It will be annoying for people like me who have to maintain different services and may want to keep the custom translations in the same directory without committing them.

Fixed this and added some more tests

nemesisdesign:
> Another thing, this will not work as expected because `plan.plan` needs to be msgID.

I did test this and it works. How can we work around the error?
I need to be able to translate that part which is defined in the DB but making the content in the DB translatable is really overkill and doing it here is going to be far faster to implement.

internals/generators/organization/config.yml.hbs
```diff
@@ -249,6 +248,11 @@ client:
         username:
           type: "email"
           pattern: '.+@.+\..+'
+          pattern: '.+@.+\..+'
+          label:
+            en: "email"
+          placeholder:
+            en: "enter email"
```
nemesisdesign, line 255: I guess we can remove this since it's already present in the translations right?

org-configurations/default-configuration.yml
```diff
@@ -181,6 +181,10 @@ client:
         username:
           type: "email"
           pattern: '.+@.+\..+'
+          label:
+            en: "email"
+          placeholder:
+            en: "enter email"
```
nemesisdesign, line 187: same here

client/utils/load-translation.js
```diff
@@ -0,0 +1,89 @@
...
+    navigator.languages.every((browserLang) => {
+      if (availLangs.includes(browserLang)) {
+        lang = browserLang;
+        return false;
+      }
+      if (availLangs.includes(browserLang.substr(0, 2))) {
+        lang = browserLang.substr(0, 2);
+        return false;
+      }
+      return true;
```
nemesisdesign, line 46: does this `true` and do the above `false` values returned here make any difference?
It's a bit confusing.

client/utils/load-translation.js
```diff
@@ -0,0 +1,89 @@
...
+      try {
+        await importTranslation(orgSlug, language);
+      } catch (error) {
+        // Using default language of that organization.
+        console.log(
+          "Cannot found translation for this language. Using default language.",
+        );
+        await importTranslation(
+          orgSlug,
+          localStorage.getItem(`${orgSlug}-defaultLanguage`),
```
nemesisdesign, line 64: why should we get this from localStorage? Don't we have it in the conf?

client/utils/load-translation.js
codesankalp: It's just to break the iteration.

org-configurations/default-configuration.yml
codesankalp: I added it here because if the user wants a custom label then he can add it for the username field.

client/utils/load-translation.js
codesankalp: Ok, I will change this to not to use local storage.

client/utils/load-translation.js
nemesisdesign: then you should be able to simply use `return`, the last 2 return statements should not be needed, or is the linter complaining? Alternatively just use `else if` insead of the second `if` in which case you should be able to avoid the return completely and no ambiguity is left.

nemesisdesign (changes requested):
Did another round of testing of the preferred browser language but it's not working as expected.

I have an org with two languages:

- Italian (default)
- English

My browser is set to prefer English.

The code does recognize this and the line `await importTranslation(orgSlug, language);` is executed without errors, however, the language on the page stays in Italian, the correct behavior should be to automatically switch to the preferred language of the user, which in this case is English.

codesankalp:
> Did another round of testing of the preferred browser language but it's not working as expected.
> 
> I have an org with two languages:
> 
> * Italian (default)
> * English
> 
> My browser is set to prefer English.
> 
> The code does recognize this and the line `await importTranslation(orgSlug, language);` is executed without errors, however, the language on the page stays in Italian, the correct behavior should be to automatically switch to the preferred language of the user, which in this case is English.

Have you tried this after clearing localStorage? I have tested the same scenario and English is being loaded in my case.

internals/generators/organization/config.yml.hbs
codesankalp: https://github.com/openwisp/openwisp-wifi-login-pages/pull/287#discussion_r671595861

client/components/registration/registration.js
```diff
@@ -259,6 +259,7 @@ export default class Registration extends React.Component {
   };
 
   getLabelText = (plan) => {
+    /* disable ttag */
```
nemesisdesign, line 262: nice :+1: 
is this to disable the warning?

codesankalp: Yes

nemesisdesign (approved):
I found a way to improve/simplify the translations-update command and took advantage to update and improve the docs, please have a look at my last 2 commits.

The rest looks good, the main requirements are being met so I think we can merge this and move on, I will keep testing it since I will be adding translations for several languages, I will surely let you know if I find any bug.

codesankalp:
> I found a way to improve/simplify the translations-update command and took advantage to update and improve the docs, please have a look at my last 2 commits.

Yes, it looks good. I don't know about this command :sweat_smile:
 
> The rest looks good, the main requirements are being met so I think we can merge this and move on, I will keep testing it since I will be adding translations for several languages, I will surely let you know if I find any bug.

yes :smile:

PR #303, opened 2021-07-07
[feature] Added support for default organization and reduce boilerplate #236

### LINKED ISSUE #59: [wifi-login-pages] Avoid the need of repeating URLs for each organization
These URLs should no be repeated for each organization:
```
server:
  host: "http://localhost:8000"
  proxy_urls:
    password_change: "/api/v1/{org_slug}/account/password/change"
    password_reset: "/api/v1/{org_slug}/account/password/reset"
    password_reset_confirm: "/api/v1/{org_slug}/account/password/reset/confirm"
    registration: "/api/v1/{org_slug}/account"
    user_auth_token: "/api/v1/{org_slug}/account/token"
    validate_auth_token: "/api/v1/{org_slug}/account/token/validate"
    authorize: "/api/v1/authorize"
```

The should be stored elsewhere in the code, not in the YAML configuration of each organization.

### LINKED ISSUE #236: [Feature] Add support for default configuration and reduce boilerplate

### LINKED ISSUE #313: [change] Refactor org-configurations boilerplate
* Change directory name from `org-configurations` to `organizations` or `orgs`.
* Change name from `{slug}-configuration.yml` to `{slug}.yml` and backward compatibility for this change.
* default configuration hardcoded in the repository and visible in prod: when the application is deployed to prod, the default configuration is visible, ideally, we should have this appear only in the development.

codesankalp:
Closes #236 
Closes #59 
Closes #313


nemesisdesign (changes requested):
I see this is going in the right direction.

### Testing the generation of a new config

I tried generating a new conf and this is the result I got:

```yaml
name: NewConf
slug: newconf
server:
  uuid: 60050402-59c9-4933-a502-8371b4aa97b6
  secret_key: MBrO9VOAHFiSicKnoAbyRUlCdvQLFShk
  proxy_urls:
    create_mobile_phone_token: '/api/v1/{org_slug}/account/phone/token'
    verify_mobile_phone_token: '/api/v1/{org_slug}/account/phone/verify'
    mobile_phone_number_change: '/api/v1/{org_slug}/account/phone/change'
client:
  components:
    header:
      logo:
        alternate_text: NewConf
      links:
        '0':
          url: /newconf/login
        '1':
          url: /newconf/registration
        '2':
          text: Status
          url: /newconf/status
        '3':
          text: Change password
          url: /newconf/change-password
          authenticated: true
    captive_portal_login_form:
      additional_fields:
        '1':
          value: 'http://localhost/newconf/status'
    captive_portal_logout_form:
      action: 'http://localhost:8000/captive-portal-mock/login/'
      logout_by_session: false
    registration_form:
      links:
        forget_password: true
    mobile_phone_verification_form:
      input_fields:
        code:
          type: text
          pattern: '[0-9]{6,}'
    phone_number_change_form:
      buttons:
        change_phone_number: true
        cancel: true
```

What are those `'0'`, `'1'`, etc? Can we keep the same list structure of the previous format?

What about: `create_mobile_phone_token`, `verify_mobile_phone_token`, `registration_form`, `mobile_phone_verification_form` etc?
These seems like we don't really need to have them repeated in every conf, right?

### Keeping only the new generated config gives me errors

I tried removing all the configs and keep only the new generated configuration, but this results in the following JS errors logged in the browser console:

```
Uncaught Error: When called with an action of type "SET_LANGUAGE", the slice reducer for key "language" returned undefined. To ignore an action, you must explicitly return the previous state. If you want this reducer to hold no value, you can return null instead of undefined.
    at combination (redux.js:532)
    at dispatch (redux.js:288)
    at index.js:11
    at dispatch (redux.js:659)
    at set-organization.js:26
    at index.js:8
    at setOrganization (index.js:18)
    at new OrganizationWrapper (organization-wrapper.js:43)
    at constructClassInstance (react-dom.development.js:12880)
    at updateClassComponent (react-dom.development.js:17100)
react-dom.development.js:19527 The above error occurred in the <OrganizationWrapper> component:
    in OrganizationWrapper (created by Connect(OrganizationWrapper))
    in Connect(OrganizationWrapper) (created by withCookies(Connect(OrganizationWrapper)))
    in withCookies(Connect(OrganizationWrapper)) (created by Context.Consumer)
    in withCookies(Connect(OrganizationWrapper)) (created by Context.Consumer)
    in Route (created by Routes)
    in Switch (created by Routes)
    in Routes (created by Connect(Routes))
    in Connect(Routes) (created by Context.Consumer)
    in Route (created by BaseApp)
    in Router (created by BaseApp)
    in BaseApp (created by Connect(BaseApp))
    in Connect(BaseApp)
    in Provider
    in CookiesProvider

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
logCapturedError @ react-dom.development.js:19527
react-dom.development.js:11102 Uncaught Error: When called with an action of type "SET_LANGUAGE", the slice reducer for key "language" returned undefined. To ignore an action, you must explicitly return the previous state. If you want this reducer to hold no value, you can return null instead of undefined.
    at combination (redux.js:532)
    at dispatch (redux.js:288)
    at index.js:11
    at dispatch (redux.js:659)
    at set-organization.js:26
    at index.js:8
    at setOrganization (index.js:18)
    at new OrganizationWrapper (organization-wrapper.js:43)
    at constructClassInstance (react-dom.development.js:12880)
    at updateClassComponent (react-dom.development.js:17100)
```

### Default values

Bringing back the `default-configuration.yml` file fixes the problem mentioned previously, but I think the default values should be stored elsewhere, in the application code, not in folders which the user is supposed to have the freedom to edit at their own will.

Can you confirm where are the defaults stored?
How can we make this file accessible to the user wanting to customize their conf?
Can we keep the default values somewhere in YAML format in order to use it as a reference which we can cite from the README?
I believe it should be a place inside the app code which the user is not supposed to edit.

### Removing things

I tried to remove social login buttons but it's not working, have you already worked on this or not yet?

Eg:

```
name: NewConf
slug: newconf
server:
  uuid: 60050402-59c9-4933-a502-8371b4aa97b6
  secret_key: MBrO9VOAHFiSicKnoAbyRUlCdvQLFShk
  proxy_urls:
    create_mobile_phone_token: '/api/v1/{org_slug}/account/phone/token'
    verify_mobile_phone_token: '/api/v1/{org_slug}/account/phone/verify'
    mobile_phone_number_change: '/api/v1/{org_slug}/account/phone/change'
client:
  components:
    header:
      logo:
        alternate_text: NewConf
      links:
        '0':
          url: /newconf/login
        '1':
          url: /newconf/registration
        '2':
          text: Status
          url: /newconf/status
        '3':
          text: Change password
          url: /newconf/change-password
          authenticated: true
    captive_portal_login_form:
      additional_fields:
        '1':
          value: 'http://localhost/newconf/status'
    captive_portal_logout_form:
      action: 'http://localhost:8000/captive-portal-mock/login/'
      logout_by_session: false
    registration_form:
      links:
        forget_password: true
    mobile_phone_verification_form:
      input_fields:
        code:
          type: text
          pattern: '[0-9]{6,}'
    phone_number_change_form:
      buttons:
        change_phone_number: true
        cancel: true
    login_form:
      social_login:
        links: []
```

codesankalp:
> Bringing back the `default-configuration.yml` file fixes the problem mentioned previously, but I think the default values should be stored elsewhere, in the application code, not in folders which the user is supposed to have the freedom to edit at their own will.

Ok, I will create a different config file.

codesankalp:
> What are those `'0'`, `'1'`, etc? Can we keep the same list structure of the previous format?

This is not resolved yet.

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+    password_reset_confirm: "/api/v1/{org_slug}/account/password/reset/confirm"
+    registration: "/api/v1/{org_slug}/account"
+    user_auth_token: "/api/v1/{org_slug}/account/token"
+    validate_auth_token: "/api/v1/{org_slug}/account/token/validate"
+    authorize: "/api/v1/authorize"
+    user_radius_sessions: "/api/v1/{org_slug}/account/session"
+    plans: '/api/v1/{org_slug}/plans'
+    create_mobile_phone_token: '/api/v1/{org_slug}/account/phone/token'
+    verify_mobile_phone_token: '/api/v1/{org_slug}/account/phone/verify'
+    mobile_phone_number_change: '/api/v1/{org_slug}/account/phone/change'
```
nemesisdesign, line 25: one of the goals is to avoid this list of URLs.
According to a comment from some time ago, this should not be necessary: https://github.com/openwisp/openwisp-wifi-login-pages/issues/59

The ideal solution is to have these URLs defined only once for all organizations, without the need of redefining these URLs in each configuration of each organization, not only in the YAML, but I think we can avoid to having them defined in the config.json as well because that occupies space and needs to be downloaded, it doesn't make sense to repeat this over and over because it's always going to be the same.

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+  default_language: "en"
+  components:
+    header:
+      logo:
+        url: "openwisp-logo-black.svg" # logo url
+        alternate_text: "openwisp"
+      links:
+        - text:
+            en: "Sign In"
+          url: "/default/login"
```
nemesisdesign, line 55: can we find a way to use `{org_slug}` here too? This would allow copying this part between orgs easier.

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+          authenticated: false
+        - text:
+            en: "Sign Up"
+          url: "/default/registration"
+          authenticated: false
+        - text:
+            en: "Change password"
+          url: "/default/change-password"
+          authenticated: true
```
nemesisdesign, line 64: looks like this link is missing: https://github.com/openwisp/openwisp-wifi-login-pages/blob/7b74c90fa4960b53cd1776f476f138fbb0554fff/internals/generators/organization/config.yml.hbs#L86-L92

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+        - alt:
+            en: "facebook"
+          icon: "facebook.svg"
+          url: "https://facebook.com/openwisp"
+
+    status_page:
+      links:
+        - text:
+            en: "Change your password"
+          url: "/{orgSlug}/change-password/"
```
nemesisdesign, line 167: I noticed that here it's named `{orgSlug}`, but above `{org_slug}`, do you know why?

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+      additional_text: true
+      input_fields:
+        password:
+          type: "password"
+          pattern: ".{6,}"
+        password_confirm:
+          type: "password"
+          pattern: ".{6,}"
+      contact_text: true
+      login_page_link: true
```
nemesisdesign, line 148: I guess we should be able to avoid all of this block

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+        login: true
+
+    password_reset_form:
+      additional_text: true
+      input_fields:
+        email:
+          type: "email"
+          pattern: '.+@.+\..+'
+      contact_text: true
+      login_page_link: true
```
nemesisdesign, line 136: I guess we should be able to avoid all of this block

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+        street:
+          type: "text"
+        city:
+          type: "text"
+        tax_number:
+          type: "text"
+          pattern: '[a-zA-Z@.+\-_\d]{1,150}'
+      additional_info_text: true
+      links:
+        login: true
```
nemesisdesign, line 127: I guess we should be able to avoid all of this block

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+          type: "password"
+          pattern: ".{6,}"
+        remember_me:
+          type: "checkbox"
+          value: true
+      additional_info_text: true
+      buttons:
+        register: true
+      links:
+        forget_password: true
```
nemesisdesign, line 198: I guess we can live without input_fields, additional_info_text, buttons, links

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+
+    password_change_form:
+      input_fields:
+        password1:
+          type: "password"
+          pattern: ".{6,}"
+        password2:
+          type: "password"
+          pattern: ".{6,}"
```
nemesisdesign, line 207: same here

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+
+    404_page:
+      heading: true
+      sub_heading: true
+      message: true
+      homepage_link: true
```
nemesisdesign, line 213: same here

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+      fields:
+        id: logout_id
+      additional_fields: []
+      logout_by_session: true
+    
+    mobile_phone_verification_form:
+      input_fields:
+        code:
+          type: "text"
+          pattern: '[0-9]{6,}'
```
nemesisdesign, line 242: same here

internals/config/default.yml
```diff
@@ -0,0 +1,252 @@
...
+
+    phone_number_change_form:
+      buttons:
+        change_phone_number: true
+        cancel: true
```
nemesisdesign, line 247: same here

README.md
```diff
@@ -252,6 +252,28 @@ Run it with:
 yarn stats
 ```
 
+### Organization Configuration
+
+Organization configuration can be generated by running:
+`yarn add-org`
+
+Organization configuration is generated from the template present in `internals/generators`.
+Default configuration is present in `internals/config`. If the configuration file of a specific
```
nemesisdesign, line 261: I'd specify `/internals/generators/config.yml.hbs` just for the sake of explicitness..

README.md
```diff
@@ -252,6 +252,28 @@ Run it with:
 yarn stats
 ```
 
+### Organization Configuration
```
nemesisdesign, line 255: maybe this section and "settings" can be merged?

nemesisdesign (commented):
Let's try to simplify the default config that gets generated. The suggestions below apply to the generator as well.

Can you please add a section in the readme for each block we remove from the default config?

internals/config/default.yml
codesankalp: Because it is replaced in the client code.

server/controllers/mobile-phone-number-change-controller.js
```diff
@@ -6,6 +7,7 @@ import qs from "qs";
 import config from "../config.json";
 import defaultConfig from "../utils/default-config";
 import Logger from "../utils/logger";
+import {mobile_phone_number_change} from "../utils/proxyURLs";
```
nemesisdesign, line 10: `proxyURLs` is inconsisent naming, please be consistent and change it to `proxy-urls`

server/utils/proxyURLs.js
```diff
@@ -0,0 +1,22 @@
...
+export const authorize = `/api/v1/authorize`;
+export const user_radius_sessions = (org_slug) =>
+  `/api/v1/${org_slug}/account/session`;
+export const plans = (org_slug) => `/api/v1/${org_slug}/plans`;
+export const create_mobile_phone_token = (org_slug) =>
+  `/api/v1/${org_slug}/account/phone/token`;
+export const verify_mobile_phone_token = (org_slug) =>
+  `/api/v1/${org_slug}/account/phone/verify`;
+export const mobile_phone_number_change = (org_slug) =>
+  `/api/v1/${org_slug}/account/phone/change`;
```
nemesisdesign, line 22: There's a lot of repetition here.
What about having a single function that reads an object instead?

Eg:

```javascript
const prefix = '/api/v1/{orgSlug}/';
const paths = {
   password_change: 'account/password/change',
   password_reset: 'account/password/reset',
   // ... etc
}

const reverse = (name, orgSlug) => {
  const path = paths[name];
  if (!path) {
    // throw error
  }
  `${prefix.replace('{orgSlug}', orgSlug)}${path}`;
}
```

You can then import only 1 function (reverse) in each file and call it like `reverse('password_change', orgSlug)`.

nemesisdesign (changes requested):
good progress @codesankalp, see below.

internals/config/default.yml
```diff
@@ -0,0 +1,244 @@
...
+      helpdesk: "+789 948 564"
+      social_links:
+        - alt:
+            en: "twitter"
+          icon: "twitter.svg"
+          url: "https://twitter.com/openwisp"
+        - alt:
+            en: "facebook"
+          icon: "facebook.svg"
+          url: "https://facebook.com/openwisp"
```
nemesisdesign, line 153: I have reasoned a bit about these default configurations and I thought the following:

these default config are going to be downloaded anyway by the user right?
it doesn't make sense to add some details like these ones in the default.
I think we should keep most of these details empty (empty lists or empty strings), while we should keep these in the generator because it will serve as an example, then the user can remove the sections they don't need. It would me a lot more practical than having to set null to each option not needed. It should result also in smaller size.
What do you think?

internals/config/default.yml
```diff
@@ -0,0 +1,244 @@
...
+        - text:
+            en: "link-2"
+          url: "https://link-2.com"
+          authenticated: true
+          verified: false
+        - text:
+            en: "link-3"
+          url: "https://link-3.com"
+          authenticated: true
+          verified: true
```
nemesisdesign, line 73: same applies to these links

internals/config/default.yml
```diff
@@ -0,0 +1,244 @@
...
+          authenticated: false
+        - text:
+            en: "Change password"
+          url: "/{orgSlug}/change-password"
+          authenticated: true
+        - text:
+            en: "Change phone number"
+          url: "/{orgSlug}/change-phone-number"
+          authenticated: true
+          method: "mobile_phone"
```
nemesisdesign, line 56: this link should be removed because it doesn't work when mobile verification is not enabled, but it's shown anyway, which is not good.

I think we could even go as far as have zero links by default.

internals/config/default.yml
```diff
@@ -0,0 +1,244 @@
...
+            url: "#facebook"
+            icon: "facebook.svg"
+          - text:
+              en: "Login with Google"
+            icon: "google.svg"
+            url: "#google"
+          - text:
+              en: "Login with Twitter"
+            icon: "twitter.svg"
+            url: "#twitter"
```
nemesisdesign, line 175: same here, we could have zero links by default

internals/config/default.yml
```diff
@@ -0,0 +1,244 @@
...
+
+    phone_number_change_form:
+      buttons:
+        change_phone_number: true
+        cancel: true
+
+  privacy_policy:
+    en: "privacy-en.md"
+  terms_and_conditions:
+    en: "terms-en.md"
```
nemesisdesign, line 244: always on the concern of size, we could change these files to have only a few words in them because they're not going to be used anyway.

README.md
```diff
@@ -254,6 +254,28 @@ yarn stats
 
 ### Settings
 
+#### Organization Configuration
+
+Organization configuration can be generated by running:
```
nemesisdesign, line 259: I think we should merge this section with https://github.com/openwisp/openwisp-wifi-login-pages/blob/e5792c332c4758c5391cbcbbf6366c2f1aa90b72/README.md#add-organization-configuration.

README.md
```diff
@@ -254,6 +254,28 @@ yarn stats
...
+`yarn add-org`
+
+Organization configuration is generated from the template present in `internals/generators/config.yml.hbs`.
```
nemesisdesign, line 262: BTW, why aren't you using articles? I would change this to "The configuration of organizations is generated ...".

README.md
```diff
@@ -254,6 +254,28 @@ yarn stats
...
+Default configuration is present in `internals/config/default.yml`. If the configuration file of a specific
+organization misses a piece of configuration then this default configuration is used to generate
+complete configuration.
+
+To remove the specific section of the configuration, you can use the `null` keyword to delete
```
nemesisdesign, line 267: The part on removing sections can be kept here (with a dedicated heading) but I guess that if we go with the proposed changes it will have to be updated.

README.md
```diff
@@ -254,6 +254,28 @@ yarn stats
 
 ### Settings
 
+#### Organization Configuration
+
+Organization configuration can be generated by running:
+`yarn add-org`
+
+Organization configuration is generated from the template present in `internals/generators/config.yml.hbs`.
+Default configuration is present in `internals/config/default.yml`. If the configuration file of a specific
```
nemesisdesign, line 263: also here: Default configuration > The default configuration

nemesisdesign (changes requested):
I tried logging in and I see that the status link points to the change password page. Very weird! Do you see that?
May be a problem with the merging?
It's interesting to see that no tests catches this, it would be good to add a failing test for this.

codesankalp:
> I tried logging in and I see that the status link points to the change password page. Very weird! Do you see that?
> May be a problem with the merging?
> It's interesting to see that no tests catches this, it would be good to add a failing test for this.

I can't replicate this.

internals/config/default.yml
codesankalp: Should I change the content of these files? 
Also, the whole content is present in the configuration, Can we use express to serve these files' content?

nemesisdesign (commented):
> > I tried logging in and I see that the status link points to the change password page. Very weird! Do you see that?
> > May be a problem with the merging?
> > It's interesting to see that no tests catches this, it would be good to add a failing test for this.
> 
> I can't replicate this.

Try with these changes:

```
diff --git a/org-configurations/default-configuration.yml b/org-configurations/default-configuration.yml
index dc53401..6417e6b 100644
--- a/org-configurations/default-configuration.yml
+++ b/org-configurations/default-configuration.yml
@@ -45,6 +45,10 @@ client:
             en: "Sign Up"
           url: "/{orgSlug}/registration"
           authenticated: false
+        - text:
+            en: "Status"
+          url: "/default/status"
+          authenticated: true
         - text:
             en: "Change password"
           url: "/{orgSlug}/change-password"
```

codesankalp:
> > > I tried logging in and I see that the status link points to the change password page. Very weird! Do you see that?
> > > May be a problem with the merging?
> > > It's interesting to see that no tests catches this, it would be good to add a failing test for this.
> > 
> > 
> > I can't replicate this.
> 
> Try with these changes:
> 
> ```diff
> diff --git a/org-configurations/default-configuration.yml b/org-configurations/default-configuration.yml
> index dc53401..6417e6b 100644
> --- a/org-configurations/default-configuration.yml
> +++ b/org-configurations/default-configuration.yml
> @@ -45,6 +45,10 @@ client:
>              en: "Sign Up"
>            url: "/{orgSlug}/registration"
>            authenticated: false
> +        - text:
> +            en: "Status"
> +          url: "/default/status"
> +          authenticated: true
>          - text:
>              en: "Change password"
>            url: "/{orgSlug}/change-password"
> ```

test-2021-07-30_01 53 45

README.md
```diff
@@ -176,6 +181,20 @@ yarn start
...
 
+#### Removing sections of configuration
+
+To remove the specific section of the configuration, you can use the `null` keyword to delete
+that field or section during the building process. For example, to remove social login links, use
+
+```yaml
+login_form:
+  social_login:
+    links: null
```
nemesisdesign, line 192: are you sure this section is still relevant after the latest changes?

internals/config/default.yml
```diff
@@ -0,0 +1,193 @@
...
+      links: []
+
+    footer:
+      links: []
+      secondary_text: false
+
+    registration_form:
+      input_fields:
+        phone_number:
+          type: "tel"
```
nemesisdesign, line 48: can we please eliminate all `type` attributes of all inputs?
This thing is only adding overhead and increasing the size of configurations with no benefits, so I think we can get rid of it.

internals/config/default.yml
```diff
@@ -0,0 +1,193 @@
...
+          pattern: ".{6,}"
+        password_confirm:
+          type: "password"
+          pattern: ".{6,}"
+      contact_text: true
+      login_page_link: true
+
+    contact_page:
+      email: "support@openwisp.co"
+      helpdesk: "+789 948 564"
```
nemesisdesign, line 113: let's have null here as default please (both)

internals/config/default.yml
nemesisdesign: > Should I change the content of these files?
Yes.

> Also, the whole content is present in the configuration, Can we use express to serve these files' content?

Definitely! That's a great idea! :+1:

nemesisdesign (changes requested):
Good progress, see below. Will continue testing.

nemesisdesign (commented):
what about org-configurations and filenames like <slug>-organization.yml, are you going to address that as well?

README.md
codesankalp: No, I am changing it.

internals/config/default.yml
codesankalp: Do I have to remove type attributes from the whole code or just from the `default.yml` file?

nemesisdesign: we can hardcode the type attributes of form inputs in the HTML and remove any reference from them in default.yml, the YAML generator and everything else.

The same could apply for any other property which does not make sense to allow to be configurable, for example: why should we allow to change the type attribute of the remember me checkbox? If you find any other occurrence of the same concept let me know.

nemesisdesign (changes requested):
looks good, I think there will be improvements to make but it's better to create new issues and work on them in parallel with the last refinements.

Please can you squash the commits in this PR so we have 1 commit for each issue closed?
If there is any unit of work not related to any issue, it's ok to keep a commit for that if it makes sense.

Thanks

nemesisdesign (approved):

PR #307, opened 2021-07-08
[feature] Implemented SAML logout

nemesisdesign:
Implemented SAML logout.

### DIFF
--- a/client/components/login/login.js
+++ b/client/components/login/login.js
@@ -65,7 +65,7 @@ export default class Login extends React.Component {
       const loginMethod = getParameterByName("login_method");
       if (loginMethod) {
         // we have to use localStorage because the page may be
-        // reloaded and the information may be lost
+        // closed and the information may be lost if we use sessionStorage
         localStorage.setItem(`${orgSlug}_logout_method`, loginMethod);
       }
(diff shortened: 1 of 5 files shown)


client/components/login/login.js
```diff
@@ -62,10 +62,21 @@ export default class Login extends React.Component {
 
     // social login / SAML login
     if (username && token) {
+      const loginMethod = getParameterByName("login_method");
+      if (loginMethod) {
+        // we have to use localStorage because the page may be
+        // reloaded and the information may be lost
+        localStorage.setItem(`${orgSlug}_logout_method`, loginMethod);
```
codesankalp, line 69: sessionStorage will work also.

codesankalp (approved):

client/components/login/login.js
nemesisdesign: > sessionStorage will work also if it's only for retrieving information when the page is reloaded.

The user may close the window (especially if it's the minibrowser which automatically pops up) and load it again from another browser.

nemesisdesign: PS: I will update the comment to reflect this.

PR #315, opened 2021-07-23
[change] Avoided loading config of all organizations #238

### LINKED ISSUE #238: [change] Avoid loading config of all organizations
I noticed that the config for all organizations is included in the webpack build for each organization, which is bad, because it means:

- the app will consume unnecessary bandwidth and memory if many organizations are used
- configurations of other organizations is leaked

We should find a solution for this.

@codesankalp how do you think we can solve this?

Related to #236.

codesankalp:
Used dynamic import of config files.

Closes #238

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,8 @@ yarn-error.log*
 
 #other
 /org-configurations
-client/config.json
+client/configs/
+client/organizations.json
 !client/assets/
 client/assets/*
 !client/assets/eye.svg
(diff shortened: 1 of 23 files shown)


codesankalp:
WIP

nemesisdesign (approved):
This one looks good, also basic testing works, I suggest merging this branch with https://github.com/openwisp/openwisp-wifi-login-pages/pull/303#pullrequestreview-717583325 so we can test all these changes together and find out if there's any interaction, but I suggest doing this only once the issues raised in the review on #303 are solved.

nemesisdesign:
Changed my mind and merged it now, let's move on!

PR #318, opened 2021-07-27
[change] Call getUserRadiusSession when userData is loaded #288

### LINKED ISSUE #288: [bug] Key is not unique in session table of status component
image

While reloading the status page, I can see the above error.

The pastSession array seems to consist of the same session object two times.

codesankalp:
Closes #288


nemesisdesign (commented):
Looks good, will double check later on, I'll focus on the config PRs for now.

PR #340, opened 2021-08-06
[ux] Improvements to reveal password feature #323

### LINKED ISSUE #323: [ux] Improvements to reveal password feature
I noticed 2 areas for improvement in this feature:

1. when there are two password fields (eg: registration, change password), it would be better if the action acted on both fields and not just one
2. when the button is clicked, it would be better to trigger the focus again on the input field, because most likely the user wants to go back to typing the password

codesankalp:
Used another reference as second ref to toggle the fields and focus
on the selected one.

Closes #323


nemesisdesign (changes requested):
The second password is revealed but the icon doesn't change accordingly:

Screenshot from 2021-08-06 15-41-09

nemesisdesign (approved):

PR #342, opened 2021-08-10
[change] Skip billing info for zero sum payments #336

### LINKED ISSUE #336: [change] Allow to skip asking billing info
Allow to skip asking billing info for credit/debit card identity verification with zero sum payments.

The code should also be updated to work with the latest version of openwisp-subscriptions.

pandafy:
Closes #336

### DIFF
--- a/client/components/registration/registration.js
+++ b/client/components/registration/registration.js
@@ -157,22 +157,20 @@ export default class Registration extends React.Component {
       postData.plan_pricing = plan_pricing.id;
       postData.requires_payment = plan_pricing.requires_payment;
     }
-    if (
-      selected_plan !== null &&
-      plan_pricing &&
-      plan_pricing.requires_invoice === true
-    ) {
-      postData.billing_info = JSON.parse(
-        JSON.stringify({
-          tax_number,
-          street,
-          city,
-          zipcode,
-          country,
-          name: `${first_name} ${last_name}`,
-        }),
-      );
+    if (selected_plan !== null && plan_pricing) {
       postData.username = username;
+      if (plan_pricing.requires_invoice === true) {
+        postData.billing_info = JSON.parse(
+          JSON.stringify({
+            tax_number,
+            street,
+            city,
+            zipcode,
+            country,
+            name: `${first_name} ${last_name}`,
+          }),
+        );
+      }
     }
     const body = JSON.parse(JSON.stringify(postData));
     setLoading(true);
@@ -464,7 +462,7 @@ export default class Registration extends React.Component {
                         />
                       </div>
 
-                      {this.doesPlanRequireInvoice() && (
+                      {this.isPlanIdentityVerifier() && (
                         <div className="row username">
                           <label htmlFor="username">
                             {input_fields.username.label
--- a/client/components/registration/registration.js
+++ b/client/components/registration/registration.js
@@ -155,11 +155,12 @@ export default class Registration extends React.Component {
     if (selected_plan !== null) {
       plan_pricing = plans[selected_plan];
       postData.plan_pricing = plan_pricing.id;
+      postData.requires_payment = plan_pricing.requires_payment;
     }
     if (
       selected_plan !== null &&
       plan_pricing &&
-      plan_pricing.verifies_identity === true
+      plan_pricing.requires_invoice === true
     ) {
       postData.billing_info = JSON.parse(
         JSON.stringify({
@@ -311,9 +312,17 @@ export default class Registration extends React.Component {
   };
 
   isPlanIdentityVerifier = () => {
+    // If a payment is required, the plan is valid for identity verification
     const {selected_plan, plans} = this.state;
     return (
-      selected_plan !== null && plans[selected_plan].verifies_identity === true
+      selected_plan !== null && plans[selected_plan].requires_payment === true
+    );
+  };
+
+  doesPlanRequireInvoice = () => {
+    const {selected_plan, plans} = this.state;
+    return (
+      selected_plan !== null && plans[selected_plan].requires_invoice === true
     );
   };
 
@@ -455,7 +464,7 @@ export default class Registration extends React.Component {
                         />
                       </div>
 
-                      {this.isPlanIdentityVerifier() && (
+                      {this.doesPlanRequireInvoice() && (
                         <div className="row username">
                           <label htmlFor="username">
                             {input_fields.username.label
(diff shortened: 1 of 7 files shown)


client/components/registration/registration.js
```diff
@@ -299,7 +299,14 @@ export default class Registration extends React.Component {
...
+      selected_plan !== null && plans[selected_plan].requires_payment === true
+    );
+  };
+
+  isPlanRequiresInvoice = () => {
+    const {selected_plan, plans} = this.state;
+    return (
+      selected_plan !== null && plans[selected_plan].requires_invoice === true
     );
   };
```
pandafy, line 312: maybe we can get rid of `isPlanIdentityVerifier` function?

pandafy: This is the only place where it is being used and I believe it can be removed without any problem.

Just need your confirmation @nemesisdesign 

https://github.com/openwisp/openwisp-wifi-login-pages/blob/e44bedc50280418b89ba18c4f39165ecebc54b87/client/components/registration/registration.js#L367-L381

server/controllers/registration-controller.js
```diff
@@ -25,7 +25,7 @@ const registration = (req, res) => {
         delete postData.phone_number;
         postData.method = "";
       }
-      if (settings && settings.subscriptions && postData.billing_info) {
+      if (settings && settings.subscriptions) {
```
pandafy, line 28: there are only two methods to register SMS and credit/debit card. The else clause above rules out that the registration method is SMS. Only other possibility is credit/debit card, hence this condition check should be sufficient.

client/components/registration/registration.js
nemesisdesign: I think we can keep it and we can keep the name as is now (`isPlanIdentityVerifier`) with thefollowing comment:

if a payment is required, the plan is valid for identity verification.

It makes the code more readable.

server/controllers/registration-controller.js
nemesisdesign: I don't think this is correct, it's also possible to disable SMS verification and just register with email.
I believe it's necessary to add an extra parameter that tells the backend with 100% certainty this is a registration which requires a credit/debit card payment.

client/components/registration/subscriptions.test.js
```diff
@@ -125,7 +138,32 @@ describe("test subscriptions", () => {
...
+        data: plans,
+      }),
+    );
+    wrapper = initShallow(props);
+    wrapper.instance().setState({plans, selected_plan: 2, plansFetched: true});
+    expect(wrapper.find(".billing-info").length).toBe(0);
+    expect(wrapper.find("input[name='username']").length).toBe(0);
+  });
+
+  it("show billing info when both requires_payment and requires_invoice is true", () => {
```
nemesisdesign, line 157: let's use BDD style here, so "should show ..."

client/components/registration/subscriptions.test.js
```diff
@@ -125,7 +138,32 @@ describe("test subscriptions", () => {
     wrapper.instance().setState({plans, selected_plan: 0, plansFetched: true});
     expect(wrapper.find(".billing-info").length).toBe(0);
     expect(wrapper.find("input[name='username']").length).toBe(0);
-    wrapper.instance().setState({selected_plan: 1});
+  });
+
+  it("don't show billing info when requires_payment is true but requires_invoice is false", () => {
```
nemesisdesign, line 143: should not show ...

nemesisdesign (changes requested):
Looks good, will test soon, I also replied to your comments above.

See my comments below.

client/components/registration/registration.js
```diff
@@ -297,9 +298,17 @@ export default class Registration extends React.Component {
...
   isPlanIdentityVerifier = () => {
+    // If a payment is required, the plan is valid for identity verification
     const {selected_plan, plans} = this.state;
     return (
-      selected_plan !== null && plans[selected_plan].verifies_identity === true
+      selected_plan !== null && plans[selected_plan].requires_payment === true
+    );
+  };
+
+  isPlanRequiresInvoice = () => {
```
nemesisdesign, line 308: `isPlanRequiresInvoice` > `doesPlanRequireInvoice`

nemesisdesign (changes requested):

nemesisdesign (commented):
I changed the target branch to a branch I'm working on because in that branch there are changes which will be painful to rebase, it's easier to port the changes here to that branch.

Please resolve the conflict, is just an indentation issue.

nemesisdesign (changes requested):
At the moment the mobile phone should not be requested in the payment flow (there's no point in asking the phone number if we don't verify it nor do anything with it).

pandafy:
> At the moment the mobile phone should not be requested in the payment flow (there's no point in asking the phone number if we don't verify it nor do anything with it).

After rebasing on the target branch, this bug disappeared somehow. I have updated the tests to ensure that phone_number field is not rendered plan that verifies identity(requires payment)

client/components/registration/subscriptions.test.js
```diff
@@ -109,25 +122,57 @@ describe("test subscriptions", () => {
...
       }),
     );
+    props.settings.mobile_phone_verification = true;
     wrapper = initShallow(props);
     wrapper.instance().setState({plans, selected_plan: 0, plansFetched: true});
     expect(wrapper.find(".billing-info").length).toBe(0);
     expect(wrapper.find("input[name='username']").length).toBe(0);
-    wrapper.instance().setState({selected_plan: 1});
+    // phone_number field should be rendered since this plan does not verifies identity
+    expect(wrapper.find("input[name='phone_number']").length).toBe(1);
```
pandafy, line 143: This check is failing here, even after adding `props.settings.mobile_phone_verification = true;`. Can you help me with this test @nemesisdesign?

nemesisdesign (changes requested):

client/components/registration/subscriptions.test.js
codesankalp, line 143:
```suggestion
    expect(wrapper.find("t[name='phone_number']").length).toBe(1);
```
When I debugged it the wrapper is showing a custom tag (`t`) for the phone input field @pandafy, So replacing it with that passes the test.

pandafy: Thank you Sankalp! After making this change, the test suite passes. 😄

nemesisdesign (changes requested):
Now that https://github.com/openwisp/openwisp-wifi-login-pages/pull/348 has been merged,
can you please cherry-pick/rebase your commits on the current master?

nemesisdesign (approved):
Thanks! :+1:

PR #344, opened 2021-08-11
[fix] Added setUserData to called after logout #343

### LINKED ISSUE #306: [change] Ensure every logout button performs captive portal logout
Let's ensure that the log out action is performed always with the same logic which also executes the captive portal logout. We can do this by using `setUserData({...userData, mustLogout: true})` and redirect to the status component which will take care of the rest.

### LINKED ISSUE #343: [bug] Signing up shows information of previously logged in user
**Steps to replicate**

1. Setup the project with openwisp-radius and enable SMS verification in organization's radius settings
2. Register first user using mobile number. The application will redirected to SMS token verification page (This is expected, complete registration process).
3. Logout first user
4. Register second user with different mobile number and email address. 

**Actual result**
Instead of redirecting user to SMS token verification page, the user is redirected to the /status page and  information of old user is shown

The verification page does appear after a reload. 

**Expected result**
User should be redirected to SMS token verification page and should no see any information of previously logged in user.

https://user-images.githubusercontent.com/32094433/128922994-4b23e791-5730-4755-b209-61b2c727eb14.mp4

codesankalp:
If the session is active then setUserData is not called after captive
portal logout. Added setUserData with initialData.userData after
performing captive portal logout.

Fixes #343
Closes #306

### DIFF
--- a/client/utils/handle-logout.js
+++ b/client/utils/handle-logout.js
@@ -1,0 +1,29 @@
+import {toast} from "react-toastify";
+import {t} from "ttag";
+import {initialState} from "../reducers/organization";
+
+const redirectToStatus = (setUserData, userData, orgSlug) => {
+  setUserData({...userData, mustLogout: true});
+  window.location.assign(`${window.location.origin}/${orgSlug}/status`);
+};
+
+const handleLogout = (logout, cookies, orgSlug, setUserData, userData) => {
+  /*
+   * Redirecting to the status page for captive-portal logout if the
+   * method is unspecified or bank_card or the user is verified.
+   */
+  if (
+    userData.is_active === true &&
+    (userData.method === "" ||
+      userData.method === "bank_card" ||
+      userData.is_verified === true)
+  ) {
+    redirectToStatus(setUserData, userData, orgSlug);
+    return;
+  }
+  logout(cookies, orgSlug);
+  toast.error(t`ERR_OCCUR`);
+  setUserData(initialState.userData);
+};
+
+export default handleLogout;
--- a/client/components/mobile-phone-change/mobile-phone-change.test.js
+++ b/client/components/mobile-phone-change/mobile-phone-change.test.js
@@ -402,4 +402,15 @@ describe("Change Phone Number: corner cases", () => {
     wrapper = await mountComponent(props);
     expect(wrapper.find(Redirect)).toHaveLength(0);
   });
+
+  it("should validate token", async () => {
+    wrapper = await mountComponent(props);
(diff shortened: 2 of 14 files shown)


nemesisdesign (changes requested):
Can you rebase on the latest master and fix conflicts please?

nemesisdesign (commented):
@codesankalp can you please indicate if there's any tests which replicates the issue and fails without the fix? If not, could you add at least one?

client/utils/handle-logout.js
```diff
@@ -0,0 +1,24 @@
...
+  window.location.assign(`${window.location.origin}/${orgSlug}/status`);
+};
+
+const handleLogout = (logout, cookies, orgSlug, setUserData, userData) => {
+  if (
+    userData.is_active === true &&
+    (userData.method === "" ||
+      (userData.method === "mobile_phone" && userData.is_verified === true) ||
+      (userData.method === "bank_card" && userData.is_verified === false))
+  ) {
```
nemesisdesign, line 16: it's not super clear what happens here, I think we should add some comments.

I think the user should always be logged out from the captive portal if is_verified is true or if the method used is bank_card or unspecified (regardless of is_verified).

client/utils/handle-logout.js
```diff
@@ -0,0 +1,24 @@
...
+    redirectToStatus(setUserData, userData, orgSlug);
+  }
+  logout(cookies, orgSlug);
+  toast.error(t`ERR_OCCUR`);
```
nemesisdesign, line 20: why is this an error? Can you add some comments please?
And if this is an error, does it means that if the if block above is triggered that's not an error? Shouldn't a return statement be present at the end of the if block above then?

nemesisdesign (changes requested):
Thanks @codesankalp, I have a few doubts, see below.

client/utils/handle-logout.js
codesankalp: It's an error because either the token is wrong or user data is empty.

nemesisdesign (commented):
Looks better now, @pandafy could you please double check if the error you were having is fixed?

pandafy:
I confirm this solves the issue I was facing. 👍🏼

nemesisdesign (changes requested):
@codesankalp could you squash the commits so we have one per issue? Thanks.

nemesisdesign:
> I confirm this solves the issue I was facing. 👍🏼

Thank you @pandafy :clap:.

nemesisdesign (approved):

PR #352, opened 2021-08-21
[Fix] Browserlang selection bug fixed #350

### LINKED ISSUE #350: [bug] Preferred language detection broken
On the latest master, I have an org that has two languages:

- Italian (default)
- English

If I set my language preference to English and open the page with a fresh incognito window, the language displayed is Italian, but it should be English.

Something broke this functionality, we need a failing test for this to ensure this won't happen again.

codesankalp:
Added a check for prevProps.language must not be empty strings.
Force re-render of child components to update translations.

Fixes #350


codesankalp:
Also, added a failing test.

nemesisdesign (approved):
Great, thanks! :+1:

PR #355, opened 2021-08-21
[change] Removed all style tag in head

codesankalp:


nemesisdesign (commented):
This was done on purpose to ensure the loading indicator is shown as soon as possible, so I am closing this. If that part cannot be minimized without leaving it there we can live with it, thanks.

PR #356, opened 2021-08-21
[feature] Added sticky message #347

### LINKED ISSUE #347: [feature] Add possibility to configure sticky message
Add the possibility to configure a sticky message with custom HTML which is shown in all pages.

It can be used for communicating scheduled maintenance operations or any other important message which the provider of the service wants to communicate to their users.

codesankalp:
Added custom html support for sticky message

Closes #347

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,6 +398,19 @@ header:
     alternate_text: "logo2"
 ```
 
+#### Sticky message
+
+```yaml
+header:
+  sticky_msg:
+    html:
+      en: >
+        <p class="announcement">
+          This site will go in schedule maintenance
+          <b>tonight (10pm - 11pm)</b>
+        </p>
+```
+
 ##### Login page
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -441,6 +441,25 @@ contact_page:
       </div>
 ```
 
+#### Header
```
nemesisdesign, line 444: I think this should be called "Sticky message" and should be listed after "Second logo" above.

nemesisdesign (changes requested):
Good progress @codesankalp, I also have a few doubts here.

nemesisdesign (approved):
I sent some improvemens in a1121d38b0dd151a1c733b083f5c3e4bcd30ce11, please double check and let me know. LGTM now :+1:.

codesankalp (commented):
Yes it looks good

PR #360, opened 2021-08-24
[fix] Added selected language in language headers while calling API #359

### LINKED ISSUE #359: [bug] Cannot translate messages from OpenWISP RADIUS API because accept-language header is not forwarded correctly
In order to get the right translations from the responses of the OpenWISP RADIUS HTTP API, we need to include the correct accept-language header in the requests we send from the server inside this nodeJS app.

This is something I had not planned in detail before GSoC, but I think it shouldn't be hard to accomplish.

#### Step 1

First of all, we should make sure that requests from the frontend app to the backend server have the right accept-language header.
Right now I see that the accept-language header contains the preferred languages which the user set in the browser, but this may not coincide with the language which is selected in the UI.

Therefore we should make sure that the accept-language header sent to the server in API requests contains the language selected in the UI.

#### Step 2

Once step 1 is done, we simply have to ensure that the accept-language header is correctly forwarded in every API request going to OpenWISP RADIUS from the server of this app.

#### Testing info

I already started translating API responses in OpenWISP RADIUS and made a test with Italian (by hardcoding the accept-header in server code):

```javascript
axios({
  method: "post",
  headers: {
    "content-type": "application/x-www-form-urlencoded",
    "accept-language": req.headers["accept-language"]
  },
  url: `${host}${obtainTokenUrl}/`,
  timeout,
  data: qs.stringify({username, password}),
})
```

Screenshot from 2021-08-23 19-24-40

This means that here we simply need to do the steps listed above and the rest will work.

codesankalp:
Added selected language in accept-language header while calling OpenWISP RADIUS API.

Fixes #359

### DIFF
--- a/server/controllers/obtain-token-controller.js
+++ b/server/controllers/obtain-token-controller.js
@@ -27,7 +27,6 @@ const sendCookies = (username, response, conf, res) => {
 
 const obtainToken = (req, res) => {
   const reqOrg = req.params.organization;
-  console.log(req.headers);
   const validSlug = config.some((org) => {
     if (org.slug === reqOrg) {
       // merge default config and custom config
@@ -41,6 +40,7 @@ const obtainToken = (req, res) => {
         method: "post",
         headers: {
           "content-type": "application/x-www-form-urlencoded",
+          "accept-language": req.headers["accept-language"],
         },
         url: `${host}${obtainTokenUrl}/`,
         timeout,
--- a/server/controllers/obtain-token-controller.js
+++ b/server/controllers/obtain-token-controller.js
@@ -27,6 +27,7 @@ const sendCookies = (username, response, conf, res) => {
 
 const obtainToken = (req, res) => {
   const reqOrg = req.params.organization;
+  console.log(req.headers);
   const validSlug = config.some((org) => {
     if (org.slug === reqOrg) {
       // merge default config and custom config
--- a/server/controllers/plans-controller.js
+++ b/server/controllers/plans-controller.js
@@ -19,6 +19,7 @@ const plans = (req, res) => {
         method: "get",
         headers: {
           "content-type": "application/x-www-form-urlencoded",
+          "accept-language": req.headers["accept-language"],
         },
         url: `${host}${plansUrl}/`,
         timeout,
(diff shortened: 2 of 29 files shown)


codesankalp:
I didn't add `accept-language` in the header while calling validateToken and account sessions API because `getErrorText` is not used by them.

server/controllers/obtain-token-controller.js
```diff
@@ -27,6 +27,7 @@ const sendCookies = (username, response, conf, res) => {
 
 const obtainToken = (req, res) => {
   const reqOrg = req.params.organization;
+  console.log(req.headers);
```
nemesisdesign, line 30: :point_up:

nemesisdesign (changes requested):
Step 2 from the issue is missing, without that OpenWISP RADIUS will always reply in English. We need to forward the accept-language header in the server API endpoints of this app to OpenWISP RADIUS, that way django will return the translated text, here's an example for obtain token controller:

```diff
diff --git a/server/controllers/obtain-token-controller.js b/server/controllers/obtain-token-controller.js
index e494302..ddc6c08 100644
--- a/server/controllers/obtain-token-controller.js
+++ b/server/controllers/obtain-token-controller.js
@@ -41,6 +41,7 @@ const obtainToken = (req, res) => {
         method: "post",
         headers: {
           "content-type": "application/x-www-form-urlencoded",
+          "accept-language": req.headers["accept-language"]
         },
         url: `${host}${obtainTokenUrl}/`,
         timeout,
```

server/controllers/plans-controller.js
```diff
@@ -30,7 +31,7 @@ const plans = (req, res) => {
             .send(response.data);
         })
         .catch((error) => {
-          console.log(error);
+          Logger.log(error);
```
nemesisdesign, line 34: pay attention, there's another call to `Logger.error(error);` just below. Maybe we can remove this line.

nemesisdesign (changes requested):

server/controllers/plans-controller.js
codesankalp: Removed

nemesisdesign (approved):
I tested login, registraton, password reset and seem to work, thanks! :+1: 

I'll merge now and use it, if I find anything which I cannot fix myself I will let you know.

PR #361, opened 2021-08-24
[change] Reusable fuction for errors #357

### LINKED ISSUE #357: [change] Create reusable function for errors
Right now we are including a lot of code like:

```
{errors.username && (
  <div className="error">
    <span className="icon">!</span>
    <span className="text">{errors.username}</span>
  </div>
)}
``` 

Maybe we can create an utility function (not sure if it's worth it to have another component for this as it will complicate the tree a lot) for this? This would give us the following advantages:

- bring down size a bit (these lines are repated in almost all components)
- make it easy to change the style of errors

I am thinking of something like:

```javascript
fieldError (errors, key) => {
  (
    {errors[key] ? (
      <div className="error">
        <span className="icon">!</span>
        <span className="text">{errors[key]}</span>
      </div>
    ) : ''}
  );
}
```

codesankalp:
Added get-error-field to remove repeated code.

Closes #357

### DIFF
--- a/client/utils/get-error-field.js
+++ b/client/utils/get-error-field.js
@@ -1,0 +1,11 @@
+import React from "react";
+
+const getErrorField = (errors, key, nonField = false, className = "") =>
+  errors[key] ? (
+    <div className={!nonField ? `error ${className}` : "error non-field"}>
+      <span className="icon">!</span>
+      <span className="text">{errors[key]}</span>
+    </div>
+  ) : null;
+
+export default getErrorField;
--- a/client/utils/get-error-field.js
+++ b/client/utils/get-error-field.js
@@ -1,11 +1,13 @@
 import React from "react";
 
-const getErrorField = (errors, key, nonField = false, className = "") =>
-  errors[key] ? (
-    <div className={!nonField ? `error ${className}` : "error non-field"}>
+const getErrorField = (errors, key) => {
+  const errorKey = key === null ? "nonField" : key;
+  return errors[errorKey] ? (
+    <div className={key !== null ? `error ${key}` : "error non-field"}>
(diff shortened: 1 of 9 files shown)


nemesisdesign (commented):
Looks good, there are a few conflicts after mergin the other PRs, I also have one doubt below.

client/utils/get-error-field.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const getErrorField = (errors, key, nonField = false, className = "") =>
+  errors[key] ? (
+    <div className={!nonField ? `error ${className}` : "error non-field"}>
```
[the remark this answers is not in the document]
codesankalp: Resolved :smile:

client/utils/get-error-field.js
```diff
@@ -1,11 +1,13 @@
 import React from "react";
 
-const getErrorField = (errors, key, nonField = false, className = "") =>
-  errors[key] ? (
-    <div className={!nonField ? `error ${className}` : "error non-field"}>
+const getErrorField = (errors, key) => {
```
nemesisdesign, line 3: if we use `key=null`, then `getErrorField(errors, null)` can become `getErrorField(errors)`.
At this point we can wonder, why name it `getErrorField` if it can also show non field errors? Then we could simply name it `getError`, what do you think?

nemesisdesign (commented):
Looking good, see below for a further simplification.

client/utils/get-error-field.js
codesankalp: Yes

nemesisdesign (approved):
Thanks :+1:

PR #371, opened 2021-08-27
[change] Simplified organization specific customization #367

### LINKED ISSUE #367: [change] Simplify organization specific customization
Configuring and deploying organizations is a bit painful because one has to add a YAML file in one location, assets in another location, terms of services and privacy policy in another location.

We could explore whether is possible to have everything in one place, eg:

```
organizations/{orgSlug}/
    {orgSlug}.yml
    privacy-en.md
    terms-en.md
    index.css
    logo.etc
```

I didin't think about this before but now that I'm using it I know for sure this will confuse people a lot.

If the app would be able to read this single directory and do whatever it needs to do, we would make it a lot easier for users to deploy and use it.

codesankalp:
Now config, client_assets and server_assets is at a central place.

Closes #367


README.md
```diff
@@ -195,6 +195,14 @@ login_form:
...
 is required to build and compile organization configurations. Edit it at your own will.
 
+#### Different configuration for same Radius organization
+
+It is possible to create a separate configuration with the same radius organization i.e.
+create a new YAML file (let say `variant.yml`) inside `organizations/{orgSlug}/`.
+Add the changes which is different from the original configuration.
+During setup of organization a new custom organization with name `{orgSlug}-variant` will
+be created which uses the same radius APIs of the original organization with slug `{orgSlug}`.
+
```
codesankalp, line 205: I tried to add some info. Is this correct?

nemesisdesign: I pushed several improvements and updates to the README

server/controllers/mobile-phone-number-change-controller.js
```diff
@@ -14,8 +14,11 @@ const mobilePhoneNumberChange = (req, res) => {
     if (org.slug === reqOrg) {
       // merge default config and custom config
       const conf = merge(defaultConfig, org);
-      const {host} = conf;
-      const url = reverse("mobile_phone_number_change", org.slug);
+      const {host, custom, radiusSlug} = conf;
+      const url = reverse(
+        "mobile_phone_number_change",
+        custom ? radiusSlug : org.slug,
```
nemesisdesign, line 20: I see this ternary operator is used in all the sever controller, is there a chance you can find a way to avoid it? For example, by loading this value in a new property and reusing this new property in all these lines?

nemesisdesign (approved):
I tested it and it works! Thanks! :+1: :+1: 
I improved the README.
I left a suggestion to improve the code slightly below, if you can find a way to do that it would be great and if so, please send a new patch because I will be merging this one now since the changes to the structure of the files are huge and going back and forth between branches is too painful, so we should just move on and adopt this new directory structure which is a lot more usable.

nemesisdesign (changes requested):
The browser tests are failing because the path for the mobile config has changed but the scripts were not updated, I tried fixing it but I couldn't do it quickly so I left it out for now. Can you fix that please?

nemesisdesign (approved):
Thanks @codesankalp :+1:

PR #393, opened 2021-09-01
[fix] Logout success toast when logging out from verification page #375

### LINKED ISSUE #375: [bug] Logout from mobile verification generates an error notification
- register with mobile phone verification
- during verification click on log out
- toast error shows up, but I don't think it should

toast-error-logout-mobile-phone-verification

codesankalp:
Used optional parameter showLogoutToast in handleLogout function.

Fixes #375


nemesisdesign (approved):
Thanks! :+1:

PR #395, opened 2021-09-02
[change] Skip reloading CSS file on translation change and added loader

codesankalp:


client/components/organization-wrapper/organization-wrapper.js
```diff
@@ -112,208 +116,213 @@ export default class OrganizationWrapper extends React.Component {
...
+              <link
+                rel="shortcut icon"
+                type="image/x-icon"
+                href={getAssetPath(orgSlug, favicon)}
+              />
+            </Helmet>
+          ) : null}
         </>
-      ) : null;
+      );
```
codesankalp, line 325: In this the only change is rearrangement i.e. moving `cssPath` block and `favicon` block outside the translation loaded block so that re-render of the page with CSS files does not take place.

nemesisdesign (changes requested):
I did a quick test and it worked, thanks!

I have been doing some long needed changes in https://github.com/openwisp/openwisp-wifi-login-pages/pull/398 which conflict with some of your changes here, please have a look at my changes and then rebase your branch on the latest master to fix conflicts.

nemesisdesign (approved):
Very good, another one fixed, Thanks and onto the next! :+1:

PR #396, opened 2021-09-02
[fix] Used default loader while loading PP or T&C #374

### LINKED ISSUE #374: [bug] "Loading" shown while loading Privacy Policy / Terms and conditions
- open chrome developer tools
- disable cache
- open login page
- change to a non English language
- select network throttling (fast 3g or slow 3g)
- open privacy policy or terms and conditions

**Expected result:**

I'd expect to see "loading" translated in the language chosen.

**Actual result:**

The word "Loading" is shown regardless of the language.

**Proposed solution:**

We can use a loading indicator (the main loader or a loading indicator within the modal).

codesankalp:
Added loader while fetching modal content.

Fixes #374

### DIFF
--- a/client/components/modal/index.css
+++ b/client/components/modal/index.css
@@ -31,3 +31,8 @@
 .modal-container .message {
   line-height: 30px;
 }
+.load-container {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+}
--- a/client/components/modal/modal.js
+++ b/client/components/modal/modal.js
@@ -80,7 +80,9 @@ export default class Modal extends React.Component {
               &#10006;
             </Link>
             {loading ? (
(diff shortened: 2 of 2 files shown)


client/components/modal/index.css
```diff
@@ -31,3 +31,8 @@
 .modal-container .message {
   line-height: 30px;
 }
+.load-container {
```
nemesisdesign, line 34: I have found out a lot of code which defines similar classes and even a `<Loader>` component, added by different people at different time.
We should put effort in avoiding to define the same thing multiple time. I have done some work in this regard which I will be pushing soon, please follow up with this and let's ensure to define things that are reused across the entire application only once.

nemesisdesign (approved):

PR #398, opened 2021-09-02
[misc] Simplified loader styles and other minor aspects

nemesisdesign:
- [change] Removed <title> contents
  It makes more sense to leave this empty because it
  will be loaded at runtime.

- [change] Centralized loading indicator HTML & CSS
  We must avoid to redefine the HTML and CSS of the loading indicator
  in mutliple places.
  This logic was defined in:  
    - index.html
    - status
    - organization wrapper
    - model
  Now the HTML is defined in the <Loader> component, while the CSS
  is defined in the main <style> tag.

- [chores] Removed untranslated not found <title>

### DIFF
--- a/client/components/modal/index.css
+++ b/client/components/modal/index.css
@@ -1,7 +1,6 @@
 .w-900 {
   max-width: 900px;
 }
-
 .text-left {
   text-align: left;
 }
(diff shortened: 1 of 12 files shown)


nemesisdesign (commented):
@codesankalp I tested this also in a prod service, so I'm merging to help you rebase https://github.com/openwisp/openwisp-wifi-login-pages/pull/395 on these changes, if you find any issue let me know.

PR #401, opened 2021-09-03
[feature] Added capability to read messages from iframe #370

### LINKED ISSUE #370: [feature] Add capability to read messages from iframe
The captive portal login may fail in different scenarios (eg: user has reached usage limits, freeradius misconfiguration, etc), but we cannot easily read the captive portal response in the iframe because it's in a different domain.

Since generally captive portal softwares allow to define the HTML page used to display captive portal failure messages, we can use `window.postMessage`:

1. the error page on the captive portal calls something like `window.parent.postMessage(radiusReply, 'authError')`.
2. our app listens for messages and then when a message is received:
   if we're dealing with an authError we can show `radiusReply` in an toast error message and log out the user (without captive portal log out because it's not needed since there will be no active session)

This should work, I will do a bit more investigation before jumping into it.

pandafy:
Closes #370


client/components/status/status.js
```diff
@@ -75,6 +75,20 @@ export default class Status extends React.Component {
         const macaddr = searchParams.get(
           captivePortalLoginForm.macaddr_param_name,
         );
+
+        window.addEventListener("message", (event) => {
+          // For security reasons, read https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#security_concern
+          if (event.origin !== new URL(captivePortalLoginForm.action).origin) {
```
nemesisdesign, line 81: I would leave the possibility of using also URLs on the same origin as the app itself, not only on the captive portal host.

client/components/status/status.js
```diff
@@ -330,7 +345,7 @@ export default class Status extends React.Component {
       //
     }
 
-    this.finalOperations();
+    setTimeout(this.finalOperations2000);
```
nemesisdesign, line 348: looks like there's a typo here

client/components/status/status.js
```diff
@@ -75,6 +75,21 @@ export default class Status extends React.Component {
...
+          if (
+            event.origin === new URL(captivePortalLoginForm.action).origin ||
+            event.origin === window.location.origin
+          ) {
+            toast.error(event.data, {
+              onOpen: () => toast.dismiss(event.data),
+            });
+            logout(cookies, orgSlug);
+            setLoading(false);
+          }
```
nemesisdesign, line 90: if we get here, chances are high we should return to avoid the code below to be executed, did you think about this?

nemesisdesign (approved):

PR #402, opened 2021-09-04
[feature] Added support for multiple CSS files #372

### LINKED ISSUE #372: [feature] Add possibility of specifying multiple CSS files
This is useful when different configuration share the same CSS but have some customizations.

codesankalp:
List of CSS file can be added in the config.
Updated comments inside the config file.

Closes #372

### DIFF
--- a/client/components/organization-wrapper/organization-wrapper.js
+++ b/client/components/organization-wrapper/organization-wrapper.js
@@ -303,7 +303,17 @@ export default class OrganizationWrapper extends React.Component {
           ) : null}
           {cssPath && orgSlug ? (
             <Helmet>
-              <link rel="stylesheet" href={getAssetPath(orgSlug, cssPath)} />
+              {Array.isArray(cssPath) ? (
+                cssPath.map((path) => (
+                  <link
+                    rel="stylesheet"
+                    href={getAssetPath(orgSlug, path)}
+                    key={path}
+                  />
+                ))
+              ) : (
+                <link rel="stylesheet" href={getAssetPath(orgSlug, cssPath)} />
+              )}
             </Helmet>
           ) : null}
           {favicon && orgSlug ? (
@@ -349,7 +359,7 @@ OrganizationWrapper.propTypes = {
     configuration: PropTypes.shape({
       title: PropTypes.string,
       pageTitle: PropTypes.string,
-      css_path: PropTypes.string,
+      css_path: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
       slug: PropTypes.string,
       name: PropTypes.string,
       favicon: PropTypes.string,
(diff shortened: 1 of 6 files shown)


internals/config/default.yml
```diff
@@ -24,8 +24,8 @@ client:
   # path of favicon
   favicon: "favicon.png"
 
-  # path of the custom css file relative to organization's folder in
-  # assets directory.
+  # path of the custom css file or files relative to organization's 
+  # folder in assets directory.
```
nemesisdesign, line 28: can we turn this into:

```yaml
css:
  - index.css
```

internals/generators/organization/config.yml.hbs
```diff
@@ -27,8 +27,8 @@ client:
   favicon: null
 {{/if}}
 
-  # path of the custom css file relative to organization's folder in
-  # assets directory.
+  # path of the custom css file or files relative to organization's
+  # folder in assets directory.
```
nemesisdesign, line 31: can we turn this into:

```yaml
css:
  - index.css
```

organizations/default/default.yml
```diff
@@ -24,8 +24,8 @@ client:
   # path of favicon
   favicon: "favicon.png"
 
-  # path of the custom css file relative to organization's folder in
-  # assets directory.
+  # path of the custom css file or files relative to organization's 
+  # folder in assets directory.
```
nemesisdesign, line 28: can we turn this into:

```yaml
css:
  - index.css
```

client/components/organization-wrapper/organization-wrapper.js
```diff
@@ -349,7 +359,7 @@ OrganizationWrapper.propTypes = {
     configuration: PropTypes.shape({
       title: PropTypes.string,
       pageTitle: PropTypes.string,
-      css_path: PropTypes.string,
+      css_path: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
nemesisdesign, line 362: we can leave css_path as a string for backward compatibility and then add `css` to be only an array.

client/components/organization-wrapper/organization-wrapper.js
```diff
@@ -303,7 +303,17 @@ export default class OrganizationWrapper extends React.Component {
           ) : null}
           {cssPath && orgSlug ? (
             <Helmet>
-              <link rel="stylesheet" href={getAssetPath(orgSlug, cssPath)} />
```
nemesisdesign, line 306: I suggest doing some pre-processing in the beginning of this JS function above, eg:

```javascript
if (!css) css = [];  // not sure this is needed, just ensure we always have a CSS array.
if (cssPath) css.push(cssPath);
```

This way the block of code below be simplified to always expect only an array and nothing else.

nemesisdesign (changes requested):
Great work @codesankalp :+1: 

See my suggestions below, we'll also need to add a few lines in the README regarding this.

I would add a new section after [Menu items](https://github.com/openwisp/openwisp-wifi-login-pages#menu-items), called "Custom CSS files", saying something like "It's possible to specify multiple CSS files if needed", providing an example and explanining that the possibility of adding multiple files can be useful when working with [Variants](https://github.com/openwisp/openwisp-wifi-login-pages#variants-of-the-same-configuration) (linking the section will be also important to help readers figure out what we're takling about).

nemesisdesign (approved):
Thanks :+1:

PR #403, opened 2021-09-04
[fix] Logout only when unauthorized or forbidden while fetching sessions #394

### LINKED ISSUE #394: [bug] User is logged out from status for temporary connection issues
While testing I have noticed the user can be logged out while in status if the fetching of the session data fails for some reason.
I think we should log out the user only if the auth token is not valid anymore, any other issue can be ignored because it could be a temporary connection glitch (eg: WiFI may have automatically changed to another network, mobile data connection, the web server may have been reloading in that moment, etc.).

codesankalp:
Fixed user logged out when temporary connection issues.

Fixes #394

### DIFF
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -239,10 +239,16 @@ export default class Status extends React.Component {
         "link" in headers && headers.link.includes("next");
       this.setState(options);
     } catch (error) {
-      logout(cookies, orgSlug);
-      toast.error(t`ERR_OCCUR`, {
-        onOpen: () => toast.dismiss(mainToastId),
-      });
+      // logout only if unauthorized or forbidden
+      if (
+        error.response &&
+        (error.response.status === 401 || error.response.status === 403)
+      ) {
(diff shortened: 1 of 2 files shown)


client/components/status/status.js
```diff
@@ -239,10 +239,16 @@ export default class Status extends React.Component {
...
+      // logout only if unauthorized or forbidden
+      if (
+        error.response &&
+        (error.response.status === 401 || error.response.status === 403)
+      ) {
+        logout(cookies, orgSlug);
+        toast.error(t`ERR_OCCUR`, {
+          onOpen: () => toast.dismiss(mainToastId),
+        });
+      }
```
nemesisdesign, line 251: This should work, but I think this bug deserves a test, because it's very annoying and we should ensure it doesn't happen again by accident if someone in the future changes these lines and we don't catch it.

nemesisdesign (changes requested):

nemesisdesign (approved):
Thanks! :+1:

PR #412, opened 2021-09-08
[feature] Added concept of public mode #399

### LINKED ISSUE #399: [feature] Introduce concept of public mode
The application could be browsed and used also from the public internet (instead of the wifi service), the user may register beforehand to the wifi service, or they may want to change their password, mobile phone or other info while connected to the internet for whatever reason.

In this case the login and logout operations will hang because the captive portal would not be reachable from the public internet.
Also, the message "You are connected to the WiFi, you can now use the internet" do not make sense in this mode.

We can find out if we're in public mode or not with a trick like the following:

- we can prepare an HTTP endpoint on one server on the internet which replicates the same URL structure of the captive portal login 
- when connected to a captive portal, the captive portal login URL should continue being resolved normally
- when connected to the internet, the captive portal login URL should resolve to what we have prepared at step 1
- in internet mode, the captive portal login request will be made to this mock URL, this URL should return a message which informs our app that we are in internet mode
- if we are in internet mode, we should:
  - skip captive portal log out
  - avoid showing "You are connected to the WiFi, you can now use the internet"

pandafy:
Closes #399

### DIFF
--- a/README.md
+++ b/README.md
@@ -768,6 +768,40 @@ your **OpenWISP WiFi Login Pages** service.
 With the right configuration, the error messages coming from freeradius or
 the captive portal will be visible to users on **OpenWISP WiFi Login Pages**.
 
+### Allowing users to manage account from the Internet
+
+With following configuration **OpenWISP WiFi Login Pages** can allow users to
+login and manage there account even when they are not connected to the
+WiFi service.
+
+Create a small web application which can serve the endpoints entered in
+`captive_portal_login_form.action` and `captive_portal_logout_form.action`
+of organization configuration.
+
+The web application should serve the following HTML on those endpoints:
+
+```html
+<!DOCTYPE html>
+<html>
+  <body>
+    <script>
+      window.parent.postMessage(
+        {type: "internet-mode"},
+        "https://wifi-login-pages.example.com/",
+      );
+    </script>
+  </body>
+</html>
+```
+
+**Note:** Replace `https://wifi-login-pages.example.com/` with `origin` of
+your **OpenWISP WiFi Login Pages** service.
+
+Assign a subdomain to both captive portal and system running the web application
+you just created. Configure your DNS such that accessing the subdomain from the
+WiFi resolves it to captive portal while accessing it from the internet resolves
+it to the web application.
+
 ### License
 
 See [LICENSE](https://github.com/openwisp/openwisp-wifi-login-pages/blob/master/LICENSE).
--- a/README.md
+++ b/README.md
@@ -771,7 +771,7 @@ the captive portal will be visible to users on **OpenWISP WiFi Login Pages**.
 ### Allowing users to manage account from the Internet
 
 The authentication flow might hang if a user tries to access their
-account from the public internet(without connecting to the WiFi service).
+account from the public internet (without connecting to the WiFi service).
 It occurs because the **OpenWISP WiFi Login Page** waits for a response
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -768,6 +768,40 @@ your **OpenWISP WiFi Login Pages** service.
 With the right configuration, the error messages coming from freeradius or
 the captive portal will be visible to users on **OpenWISP WiFi Login Pages**.
 
+### Allowing users to manage account from the Internet
+
+With following configuration **OpenWISP WiFi Login Pages** can allow users to
```
nemesisdesign, line 773: With the

README.md
```diff
@@ -768,6 +768,40 @@ your **OpenWISP WiFi Login Pages** service.
...
+login and manage there account even when they are not connected to the
```
nemesisdesign, line 774: there > their

README.md
```diff
@@ -768,6 +768,40 @@ your **OpenWISP WiFi Login Pages** service.
...
+WiFi service.
```
nemesisdesign, line 775: I would change this text to make it clear that the problem is that the captive portal will most probably not be reachable from the public internet, so login and logout requests will hang.

Is not that the web app cannot work from the internet, it works but those requests hang making it look like it doesn't work (but if you wait long enough the request will time out and flow will continue).

README.md
```diff
@@ -768,6 +768,40 @@ your **OpenWISP WiFi Login Pages** service.
...
+</html>
+```
+
+**Note:** Replace `https://wifi-login-pages.example.com/` with `origin` of
+your **OpenWISP WiFi Login Pages** service.
+
+Assign a subdomain to both captive portal and system running the web application
+you just created. Configure your DNS such that accessing the subdomain from the
+WiFi resolves it to captive portal while accessing it from the internet resolves
+it to the web application.
```
nemesisdesign, line 803: Assign a dedicated DNS name to be used by both systems: the captive portal and the web application which simulates it. Then configure your captive portal to resolve this DNS name to its own IP, while the public DNS resolution should point to the mock app just created.
This way captive portal login and logout requests will not hang, allowing users to view/modify their account data also from the public internet.

client/components/status/status.js
```diff
@@ -388,20 +391,35 @@ export default class Status extends React.Component {
...
-      /* disable ttag */
-      toast.error(gettext(message), {autoClose: 10000});
-      /* enable ttag */
-      logout(cookies, orgSlug);
-      setLoading(false);
+      switch (type) {
+        case "authError":
+          if (!message) {
+            break;
+          }
```
nemesisdesign, line 403:
```javascript
if (!message) break;
```

may  be more compact

nemesisdesign (changes requested):
Great work @pandafy, please see my comments below and rebase on the latest master.

README.md
```diff
@@ -768,6 +768,45 @@ your **OpenWISP WiFi Login Pages** service.
 With the right configuration, the error messages coming from freeradius or
 the captive portal will be visible to users on **OpenWISP WiFi Login Pages**.
 
+### Allowing users to manage account from the Internet
+
+The authentication flow might hang if a user tries to access their
+account from the public internet(without connecting to the WiFi service).
```
nemesisdesign, line 774: there's a missing space here before the rounded parentheses

nemesisdesign (changes requested):

nemesisdesign (approved):
Great, thanks! :+1:

PR #414, opened 2021-09-09
[fix] Bold font added #408

### LINKED ISSUE #408: [bug] Bold font seems to be missing
It seems to me that Montserrat bold is missing. I wonder what this is?
https://github.com/openwisp/openwisp-wifi-login-pages/blob/da930a9d7d95f6b29b8cf2fc4d50d7068be9844b/client/assets/fonts/montserrat-semi-bold.woff2

I think we should replace that with real bold.

I tested this on chrome (linux and android), on Firefox it looks ok (I wonder why this difference).

Look at the GIF to understand what I mean.

font-wifi-login-pages

codesankalp:
Replaced semi-bold font with bold montserrat font.

Fixes #408


codesankalp:
**Firefox**:
 
image

**Chrome**:

image

nemesisdesign (approved):
Thanks!

PR #417, opened 2021-09-13
[ci] Cache python dependencies to run openwisp-radius #407

### LINKED ISSUE #407: [ci] Cache python dependencies to run openwisp-radius
At the moment I see yarn deps are cached but the python deps are not:
https://github.com/openwisp/openwisp-wifi-login-pages/blob/e23089f9d6771b149b5038d933ab342b187e3763/.github/workflows/ci.yml 

I wonder if we can find a way to cache the dependencies used to build OpenWISP RADIUS for the browser tests, as we did for the yarn packages.
It would be nice to also be able to cache the git repository to avoid cloning it each time, because this constant cloning of OpenWISP RADIUS is polluting the clone statistics, which are a useful metric to follow to estimate the usage of the project. 

These actions should speed up the browser test phase, which now takes a considerable amount of time.

codesankalp:
Removed cloning of openwisp-radius
Cache python dependencies

Closes #407

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,11 +63,6 @@ jobs:
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
           restore-keys: ${{ runner.os }}-yarn-
 
-      - name: Set up Python 3.7
-        uses: actions/setup-python@v2
-        with:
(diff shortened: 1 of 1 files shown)


nemesisdesign (approved):
Thanks!

PR #418, opened 2021-09-13
[refactor] Removed intervalId from state in status component #406

### LINKED ISSUE #406: [refactor] Avoid using state for intervalId in status component
I believe using react state for the intervalId here is not necessary:
https://github.com/openwisp/openwisp-wifi-login-pages/blob/e23089f9d6771b149b5038d933ab342b187e3763/client/components/status/status.js#L204

We should be able to use a simple instance attribute, this should avoid the extra rendering.

codesankalp:
Added this.intervalId and fixed related tests.

Closes #406

### DIFF
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -42,7 +42,6 @@ export default class Status extends React.Component {
       userInfo: {},
       currentPage: 1,
       hasMoreSessions: false,
-      intervalId: null,
       screenWidth: window.innerWidth,
       loadSpinner: true,
       modalActive: false,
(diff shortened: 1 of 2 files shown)


client/components/status/status.js
```diff
@@ -42,13 +42,13 @@ export default class Status extends React.Component {
...
       currentPage: 1,
       hasMoreSessions: false,
-      intervalId: null,
       screenWidth: window.innerWidth,
       loadSpinner: true,
       modalActive: false,
       rememberMe: false,
     };
     this.repeatLogin = false;
+    this.intervalId = null;
```
nemesisdesign, line 51: is this really needed? it should work without this, please double check, if it works, we can omit it.

nemesisdesign (approved):
Thanks!

PR #427, opened 2021-09-22
[fix] Merge array with source object array value in variants config #424

### LINKED ISSUE #424: [bug] Variants: lists should overwrite parent
**How to replicate the issue:**

Given an organization having a conf like:

```yaml
---
name: "Org"
client:
  components:
    contact_page:
      social_links:
        - url: "http://org.com/"
          text:
            en: "Org"
```

and a variant version like:

```
---
name: "Variant"
client:
  components:
    contact_page:
      social_links:
        - url: "http://url.com/"
          text:
            en: "Variant"
```

**Expected result**: `social_links` contains only a link to `url.com`.

**Actual result**: both links (parent org link and variant) are shown.

**General recommendations**: We had similar issues in netjsonconfig and OpenWISP controller, generally when users write different lists they want these to be overridden completely in 99% cases. I fixed that problem in OpenWISP controller and netjsonconfig, we should do the same here and add a brief mention in the variants README section. When lists are being merged, the variant shall overwrite completely the previous list.

codesankalp:
Added a customizer to merge array.

Fixes #424

### DIFF
--- a/config/setup.js
+++ b/config/setup.js
@@ -19,6 +19,13 @@ const serverConfigs = [];
 const organizations = [];
 const defaultConfigFile = path.join(internalConfigDir, "default.yml");
 
+// eslint-disable-next-line consistent-return
+const customizer = (objValue, srcValue) => {
+  if (_.isArray(objValue) && _.isArray(srcValue)) return srcValue;
+};
+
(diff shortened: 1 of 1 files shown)


nemesisdesign (approved):
Thanks! :+1:

PR #428, opened 2021-09-24
[feature] User can signup to multiple organizations #419

### LINKED ISSUE #419: [feature/change] Allow users to sign up to multiple organizations
Depends on https://github.com/openwisp/openwisp-radius/issues/297.

We should intercept the 409 (conflict) HTTP status code from registration, and show an error in a modal window like the one used in logout (@codesankalp can we make that modal code reusable instead of duplicating it?).

The modal should be something like (pseudocode):

```
{response.data.detail}

if response.data.organizations:
  <p>
     {t`CONFLICT_ORGS`}:
  </p>
  <ul>
    {response.data.organizations.map(function(org))} 
      <li>{org.name}</li>
    {endloop}
  </ul>

<p>
  {t`CONFLICT_SIGNIN`}
<p>

<button>{t`YES`}</button>
<button>{t`NO`}</button>
```

If the user clicks to yes it shall be taken to log in where they can enter their credentials, log in and be automatically associated to the organization they're trying to use.

codesankalp:
Showing a modal if user is already registered with other organizations.
If user clicks yes then it gets redirected to login.

Closes #419

### DIFF
--- a/client/components/registration/registration.js
+++ b/client/components/registration/registration.js
@@ -865,19 +865,21 @@ export default class Registration extends React.Component {
             toggleModal={this.toggleModal}
             handleResponse={this.handleResponse}
             content={
-              errors.organizations.length > 0 && (
-                <div className="message">
-                  <p>{t`CONFLICT_ORGS`}</p>
-                  <ul>
-                    {errors.organizations.map((org) => (
-                      <li key={org.slug} className="org-list">
-                        <a href={`/${org.slug}`}>{org.name}</a>
-                      </li>
-                    ))}
-                  </ul>
-                  <p>{t`CONFLICT_SIGNIN`}</p>
-                </div>
-              )
+              <div className="message">
+                <p>
+                  {errors.organizations.length === 0
+                    ? t`NO_ORGS`
+                    : t`CONFLICT_ORGS`}
+                </p>
+                <ul>
+                  {errors.organizations.map((org) => (
+                    <li key={org.slug} className="org-list">
+                      {org.name}
+                    </li>
+                  ))}
+                </ul>
+                <p>{t`CONFLICT_SIGNIN`}</p>
+              </div>
             }
           />
         )}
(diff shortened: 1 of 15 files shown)


i18n/en.po
```diff
@@ -583,12 +583,21 @@ msgstr ""
...
 msgid "PLEASE_LOGIN"
 msgstr ""
+
+#: client/components/registration/registration.js:870
+msgid "CONFLICT_ORGS"
+msgstr "This user is registered with other organizations:"
+
+#: client/components/registration/registration.js:878
+msgid "CONFLICT_SIGNIN"
+msgstr "Do you want sign in to this organization?"
```
nemesisdesign, line 603: Do you own the account mentioned above and want to sign in with this organization?

Choosing yes will take you to the log in form where you can log in to this service using your existing credentials.

client/components/registration/registration.js
```diff
@@ -832,12 +852,38 @@ export default class Registration extends React.Component {
...
+            toggleModal={this.toggleModal}
+            handleResponse={this.handleResponse}
+            content={
+              errors.organizations.length > 0 && (
+                <div className="message">
+                  <p>{t`CONFLICT_ORGS`}</p>
+                  <ul>
+                    {errors.organizations.map((org) => (
+                      <li key={org.slug} className="org-list">
+                        <a href={`/${org.slug}`}>{org.name}</a>
```
nemesisdesign, line 874: let's not link the org name.

client/components/registration/registration.js
```diff
@@ -24,6 +24,8 @@ import Modal from "../modal";
 import getError from "../../utils/get-error";
 import getLanguageHeaders from "../../utils/get-language-headers";
 import redirectToPayment from "../../utils/redirect-to-payment";
+import InfoModal from "../../utils/modal";
```
nemesisdesign, line 27: why are you calling this `InfoModal` here while `Modal` in the status component? I would avoid this because it makes it harder to find where this component is used, one may think these are 2 different things but it looks to me it's the same component named differently, which is not a good idea unless the component is behaving in 2 completely different ways.

nemesisdesign (changes requested):
Great progress @codesankalp, please see my comments below.

nemesisdesign (commented):
Thanks! :+1: 

Follow up work will be: https://github.com/openwisp/openwisp-radius/issues/304.

nemesisdesign (approved):

PR #430, opened 2021-09-29
[change] Added support for creating organization from variants #429

### LINKED ISSUE #429: [change] Add support for creating organization from variants
**Requirement**:
Sometimes we need another organization with the same config and the page is mostly the same. It would be very handy to handle this with variants without having to duplicate and maintain 3 different config files which will be mostly identical

**Change Required**:
to change the code so that we can override slug, UUID, and secret key in variants.

codesankalp:
If slug is different from the parent configuration then creating a new organization
with different slug during setup.

Closes #429

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,27 +266,33 @@ the example above has been kept short for brevity.
 of objects (eg: menu links), the array/list defined in the variant always
 overwrites fully what is defined in the parent configuration file.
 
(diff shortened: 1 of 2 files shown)


nemesisdesign (approved):
Great, thanks! :+1: 

I improved the docs slightly.

PR #457, opened 2021-10-26
[fix] Add support for old browsers

### LINKED ISSUE #455: [bug] App not working on old Android versions (webview)
Several issues are happening on different old browsers.

On the Android 7 webview which opens up after connecting to a WiFI protected by captive portal, localStorage is missing, therefore we can infer that also sessionStorage is missing.

On Android 4 and probably other old browsers, `Map` and `Set` are missing, which are required by react.

## Polyfills

We should enable the polyfills to ensure the app works also on old browsers, although we should still be concerned about the size of the main chunk.

Ideally we could load the polyfills only if they're needed and have them in a separate chunk, that would be best. However, we should proceed with this optimization only as last step once we are sure the main issues are fixed and the app is usable also on those browsers.

## Not relying on localStorage, sessionStorage

We could use polyfills for these, or we could check whether in some cases we could refactor the code to not use localStorage / sessionStorage, if it makes sense. I think we'll have to evaluate this case by case.

codesankalp:
Fixes #455


nemesisdesign (changes requested):
During my testing, I found out an old browser (Samsung default browser on Android 4) which still didn't work, so I added some changes which are targeted at fixing it, with this fixed I think we will be covered for the vast majority of old browsers.

It seems that the solution to avoid loading all the polyfills in the main chunk works! Thanks for that! :+1: 
I had to change the if to use `if(typeof window.fetch !== "undefined")` which is more resilient, some browsers, especially old ones, can fail as soon as they find an undefined variable.

The polyfill.io script requires a line in the README to recommend adding it to the allowed hostnames (walled garden) of the captive portal (otherwise it won't be loaded), can you add this please?

The solution for the modal window I think is to make it full width on narrow screens (and full height too for TOS and privacy policy), there's something to polish in this regard so I opened https://github.com/openwisp/openwisp-wifi-login-pages/issues/459 as a follow up which we can do once this is merged.

The only problem remained to solve are the notification toasts which cover the page and make the app unusable on these old browsers, can you please help me find a solution? Ideally we could restrict their height on these old browsers so they can't occupy the whole screen. The question would be how do we detect these old browsers? We could check whether animations/transformations are missing? Anything is good as long as it works. 

So to recap, I ask you to please take care of the following:

- [ ] **Fix react-tostify for old browsers**
- [ ] Update README to mention polyfill.io walled garden
- [ ] Review my changes and polish if needed

client/components/login/index.css
```diff
@@ -52,4 +52,7 @@ div.remember-me {
     height: 25px;
     bottom: -6px;
   }
+  .Toastify__toast-container {
+    height: 5vw;
```
nemesisdesign, line 56: are you sure that `5vw` is a value old browsers would support?

client/polyfills.js
```diff
@@ -0,0 +1,24 @@
...
+import "core-js/es/string";
+import "regenerator-runtime/runtime";
+import "raf/polyfill";
+
+const removeToastsInOldBrowsers = () => {
+  setInterval(() => {
+    const toastContainer = document.querySelector(".Toastify").children;
+    if (toastContainer.length) {
+      const toasts = Array.from(toastContainer[0].children);
+      toasts.map((toast) => setTimeout(() => toast.remove(), 1000));
```
nemesisdesign, line 15: I would let the toast to show for at least 4 seconds (the time to read it)

codesankalp: Ok will change it.

client/components/login/index.css
codesankalp: Yes, it supports. I tested it on emulator.

nemesisdesign: I tested it on samsung browser and it doesn't support it,

Can you import the CSS only if polyfill is loaded and use max-height with a fixed pixel instead? I would also try adding a white background and dark gray text color to the toast to make sure it shows with color (I see a transparent box instead), as long as this CSS is loaded only from the polyfill file we should be fine.

nemesisdesign (approved):

PR #458, opened 2021-10-27
[enhancement] Added iframe for processing payment

pandafy:


client/components/payment-process/payment-process.js
```diff
@@ -0,0 +1,115 @@
...
+      isVerified === true ||
+      !userData.payment_url
+    ) {
+      return redirectToStatus();
+    }
+
+    // likely somebody opening this page by mistake
+    if (isAuthenticated === false || this.isTokenValid === false) {
+      return redirectToStatus();
+    }
```
pandafy, line 89: In tests `this.isTokenValid` is having `undefined` value even when `validateToken` is mocked to reply with `true`.

pandafy (commented):
After these changes the following test is failing. I don't understand why 

https://github.com/openwisp/openwisp-wifi-login-pages/blob/fcd340cbd9f08ec61867750f62dde0d2d944a7d1/client/components/status/status.test.js#L793-L848

nemesisdesign:
> After these changes the following test is failing. I don't understand why
> 
> https://github.com/openwisp/openwisp-wifi-login-pages/blob/fcd340cbd9f08ec61867750f62dde0d2d944a7d1/client/components/status/status.test.js#L793-L848

redirect to another component with react doesn't call location.assign

client/components/status/status.js
```diff
@@ -172,12 +173,12 @@ export default class Status extends React.Component {
...
 
   async finalOperations() {
-    const {userData, settings} = this.props;
+    const {userData, orgSlug, settings} = this.props;
     const {setLoading} = this.context;
     // if the user needs bank card verification,
     // redirect to payment page and stop here
     if (needsVerify("bank_card", userData, settings)) {
-      window.location.assign(userData.payment_url);
+      redirectToPayment(orgSlug);
```
nemesisdesign, line 181: This should redirect to the payment process component, which should know whether to load the payment_url in an iframe or just redirect to it, depending on configuration.

nemesisdesign (commented):
As discussed, it would be great to be able to keep the current flow or use the new one based on the iframe  depending on a configuration flag, eg:

```
settings:
  subscriptions: true
  payment_iframe: true|false
```

nemesisdesign (commented):
Please also rebase on the latest master.

nemesisdesign (commented):
We'll have to handle errors from the server side gracefully (eg: show a toast error and redirect to payment failed).

nemesisdesign (commented):
we should also show the loading overlay while the iframe loads, it should be possible to remove once the load event of the iframe fires or in a similar way as we do with captive portal login / logout.

nemesisdesign (changes requested):
I can't get it to work, I am stuck in a loop after registration, this does not happen on master.
There's something wrong on this branch.
Check your email.

client/components/payment-status/payment-status.js
```diff
@@ -109,7 +109,7 @@ export default class PaymentStatus extends React.Component {
               <div className="row">
                 <Link
                   className="button full"
-                  to={`/${orgSlug}/status`}
+                  to={`/${orgSlug}/payment/process`}
```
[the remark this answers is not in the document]
nemesisdesign: I think the setting which controls the behavior requested in the previous comment here is not there yet, right?

BTW I have observed that if I go to the status component, the captive portal login is triggered.
I think we need a dedicated setting which controls whether captive portal login is done for unverified users having bank_card as registration method.

In the use case we're working, we'll set this setting to false so that cp login is not done.

client/components/payment-process/index.css
```diff
@@ -0,0 +1,10 @@
+.payment-buffer .main-column {
+  padding: 10px;
+  border-radius: 5px;
+}
+
+iframe {
+  width: 100%;
+  height: 500px;
+  border-radius: 5px;
```
nemesisdesign, line 9: I would add `border: 0 none;` to avoid making it super obvious that it's an `iframe`.

nemesisdesign (changes requested):
Good progress, I have tested it and found out some minor issues, see my comments below.

nemesisdesign (changes requested):
I need you to rebase the target branch and this branch on the latest master to include this commit: https://github.com/openwisp/openwisp-wifi-login-pages/commit/6c82a251b2863dfe1da35bc3fe3f704a6c390994.

client/components/payment-process/payment-process.js
```diff
@@ -0,0 +1,138 @@
...
+    const {userData, cookies, orgSlug, setUserData} = this.props;
+    const {setLoading} = this.context;
+    const {message, type} = event.data;
+    // For security reasons, read https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#security_concern
+    if (
+      (userData.payment_url &&
+        event.origin === new URL(userData.payment_url).origin) ||
+      event.origin === window.location.origin
+    ) {
+      if (type === "paymentClose") {
```
nemesisdesign, line 61: I think we can now change to a switch statement since the possible cases are 3 now (I added setHeight)

internals/config/default.yml
```diff
@@ -6,6 +6,10 @@ settings:
   radius_realms: false
   mobile_phone_verification: false
   subscriptions: false
+  oneTimeTokenName: "status_token"
+  payment_iframe: false
+  requires_temporary_internet: false
```
nemesisdesign, line 11: what about renaming this to `payment_requires_internet`?

internals/config/default.yml
```diff
@@ -6,6 +6,10 @@ settings:
...
+
```
nemesisdesign, line 12: let's avoid this extra blank line

client/utils/validate-token.js
```diff
@@ -17,7 +18,14 @@ const validateToken = async (
...
   // calling validate token API only if userData.radius_user_token is undefined
-  if (token && userData && userData.radius_user_token === undefined) {
+  // or payment_url of user is undefined
+  if (
+    userData &&
+    ((token && userData.radius_user_token === undefined) ||
+      (userData.method === "bank_card" &&
+        userData.is_verified !== true &&
+        !userData.payment_url))
+  ) {
```
nemesisdesign, line 28: reminder for me: this is getting quite complex...

nemesisdesign (changes requested):
I think I found a bug (hopefully minor one):

Try opening `http://localhost:8080/dev/payment/process` in an incognito window, I see an error message toast and the loading overlay, but the loading overlay is never removed.

The console says:

```
react-dom.development.js:88 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in PaymentProcess (created by Connect(PaymentProcess))
    in Connect(PaymentProcess) (created by Context.Consumer)
    in Suspense (created by Context.Consumer)
```

nemesisdesign (commented):
> I think I found a bug (hopefully minor one):
> 
> Try opening `http://localhost:8080/dev/payment/process` in an incognito window, I see an error message toast and the loading overlay, but the loading overlay is never removed.
> 
> The console says:
> 
> ```
> react-dom.development.js:88 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
>     in PaymentProcess (created by Connect(PaymentProcess))
>     in Connect(PaymentProcess) (created by Context.Consumer)
>     in Suspense (created by Context.Consumer)
> ```

The same happens to me when the max payment limit is reached.

client/components/payment-process/payment-process.js
```diff
@@ -0,0 +1,146 @@
...
+      return redirectToStatus();
+    }
+
+    if (isTokenValid === true && !settings.payment_iframe) {
+      window.location.assign(userData.payment_url);
+    }
+
+    return (
+      <div className="container content">
+        <div className="inner payment-buffer">
```
nemesisdesign, line 120: shall we rename payment-buffer to payment-iframe-container? For clarity.

nemesisdesign (commented):
It's looking good! See my comment below.

nemesisdesign (commented):
Some tests and QA checks are failing.

nemesisdesign (changes requested):
Please see my comment below.

nemesisdesign (commented):
Seems to work fine now, I'll continue stress testing it and if I don't find any other critical issue I'll merge.

nemesisdesign (approved):

PR #472, opened 2021-11-01
[chores:ux] Hide main scrollbar when modals are open #459

### LINKED ISSUE #459: [chores:ux] Hide main scrollbar when modals are open
While working to fix the modal on old browser I experimented in making it full screen on narrow screens, but it doesn't look great because the main scrollbar shows up. 

I think we can hide the main scrollbar (set `overflow:hidden` to body and html) while the modal is open, as long as these properties are added back when the modal closes.

harsh-9in:
Hide main scrollbar when modals are open
Fixes #459

### DIFF
--- a/client/components/modal/index.css
+++ b/client/components/modal/index.css
@@ -16,7 +16,7 @@
 }
 .message {
   max-height: 70vh;
-  overflow: auto;
+  overflow: hidden;
 }
 .modal-container .message h1 {
   font-size: 21px;


client/components/modal/index.css
```diff
@@ -16,7 +16,7 @@
 }
 .message {
   max-height: 70vh;
-  overflow: auto;
+  overflow: hidden;
```
codesankalp, line 19: This change doesn't fix the issue. Please see #457 on smaller and wide screen sizes and you will see why scrollbar is shown.

codesankalp (changes requested):

nemesisdesign (commented):
#457 has been merged and its branch automatically deleted, please change the target branch to master or open a new PR if needed. Please also follow @codesankalp advice.

Here's how the issue looks:

Screenshot from 2021-11-01 16-16-18

Do you see there are two scrollbars? The outer one should not be shown while the modal is open.

PR #474, opened 2021-11-02
[fix] Added redirect url for captive portal login in old browsers #470

### LINKED ISSUE #470: [bug] Old android browsers have an issue with the status page
Some old browsers do not like the fact that we do the captive portal POST request in a browser, because they cannot detect that the system is connected to the internet. Some keep showing that the system is not connected to the internet, others do not give an easy way to close the minibrowser.

For these reasons, we should allow to set up a URL in the configuration (eg: `old_browsers_redirect_url`), to redirect after successful captive portal log in.

We should add the current status URL as a querystring parameter, eg:

```javascript
redirectUrl = `${redirectUrl}?statusUrl=${encodeURI(window.location.href)}`
window.location.assign(redirectUrl)
```

codesankalp:
Fixes #470

### DIFF
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -27,6 +27,7 @@ import InfoModal from "../../utils/modal";
 import {localStorage} from "../../utils/storage";
 import history from "../../utils/history";
 import handleSession from "../../utils/session";
+import isOldBrowser from "../../utils/is-old-browser";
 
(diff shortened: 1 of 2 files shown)


nemesisdesign:
Sorry, this was a bad idea on my side to ask. It can't work because it will cause a redirect loop.

PR #475, opened 2021-11-03
[chores:ux] Hide main scrollbar when modals are open #459

### LINKED ISSUE #459: [chores:ux] Hide main scrollbar when modals are open

codesankalp:
Closes #459

### DIFF
--- a/client/components/modal/modal.js
+++ b/client/components/modal/modal.js
@@ -22,11 +22,13 @@ export default class Modal extends React.Component {
   }
 
   async componentDidMount() {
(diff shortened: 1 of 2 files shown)


nemesisdesign (approved):
Thanks :+1:

PR #488, opened 2021-12-06
[Fix] Registration fix on mini-browsers which do not have cookies #487

### LINKED ISSUE #487: [bug] App fails on mini-browsers which do not have cookies
After days of debugging I nailed down the problem to this line:

https://github.com/openwisp/openwisp-wifi-login-pages/blob/d762b179a449bc3435ab53020115501200462f87/client/utils/validate-token.js#L20

We cannot rely on cookies for anything critical and we should always have a fallback.

I tested this change which allows the registration to avoid failing: https://github.com/openwisp/openwisp-wifi-login-pages/commit/1ea5e0ab2a402d43b893a03b625c1c80700b0ea8.

However, it introduces other issues because after that the verification process does not work properly, so it needs more work, but the though part should be over, now we only need to ensure everything keeps working (automated testing and manual testing) also after the changes.

codesankalp:
Closes #487

### DIFF
--- a/client/utils/needs-verify.js
+++ b/client/utils/needs-verify.js
@@ -9,7 +9,7 @@ const needsVerify = (method, user, settings) => {
 
   if (method === "mobile_phone") {
     return (
-      user.method === "mobile_phone" &&
+      (user.method === "mobile_phone" || user.radius_user_token !== "") &&
       user.is_verified === false &&
       settings.mobile_phone_verification
     );
@@ -17,7 +17,7 @@ const needsVerify = (method, user, settings) => {
 
   if (method === "bank_card") {
     return Boolean(
-      user.method === "bank_card" &&
+      (user.method === "bank_card" || user.radius_user_token !== "") &&
         user.is_verified === false &&
         user.payment_url &&
         settings.subscriptions,
--- a/client/utils/validate-token.js
+++ b/client/utils/validate-token.js
@@ -17,13 +17,18 @@ const validateToken = async (
   language,
 ) => {
   const url = validateApiUrl(orgSlug);
-  const authToken = cookies.get(`${orgSlug}_auth_token`);
+  const authToken = userData.auth_token
+    ? userData.auth_token
+    : cookies.get(`${orgSlug}_auth_token`);
   const {token, session} = handleSession(orgSlug, authToken, cookies);
   // calling validate token API only if userData.radius_user_token is undefined
   // or payment_url of user is undefined
+  const validate = userData.is_verified
+    ? userData.radius_user_token === undefined
+    : !userData.is_verified && userData.radius_user_token === undefined;
   if (
     userData &&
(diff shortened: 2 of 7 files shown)


client/utils/validate-token.js
```diff
@@ -17,13 +17,18 @@ const validateToken = async (
   language,
 ) => {
   const url = validateApiUrl(orgSlug);
-  const authToken = cookies.get(`${orgSlug}_auth_token`);
+  const authToken = userData.auth_token
+    ? userData.auth_token
+    : cookies.get(`${orgSlug}_auth_token`);
```
nemesisdesign, line 22: what about:

```javascript
// get auth token from redux state, otherwise try getting it from cookies
const authToken = userData.auth_token || cookies.get(`${orgSlug}_auth_token`);
```

server/controllers/validate-token-controller.js
```diff
@@ -19,7 +19,7 @@ const validateToken = (req, res) => {
       const validateTokenUrl = reverse("validate_auth_token", getSlug(conf));
       const timeout = conf.timeout * 1000;
       let {token} = req.body;
-      if (req.body.session === "false") {
+      if (req.body.session === "false" && token.includes(".")) {
```
nemesisdesign, line 22: why is this needed?

client/utils/needs-verify.js
```diff
@@ -9,15 +9,15 @@ const needsVerify = (method, user, settings) => {
...
+      (user.method === "mobile_phone" || user.radius_user_token !== "") &&
       user.is_verified === false &&
       settings.mobile_phone_verification
     );
   }
 
   if (method === "bank_card") {
     return Boolean(
-      user.method === "bank_card" &&
+      (user.method === "bank_card" || user.radius_user_token !== "") &&
```
nemesisdesign, line 20: also here and above I wonder why is this needed?

nemesisdesign (commented):
Good progress @codesankalp, see my comments below.

nemesisdesign (changes requested):
During testing, I found issues:

- after successful verification, the captive portal login is opened in a new window
- when using mobile phone verification: the status page does not show up

sankalp-reg-without-cookies

server/controllers/validate-token-controller.js
codesankalp: When the cookie is used then it was signed from the express backend but if we set the token directly then it must not be signed, So, skipping the step of signing token if it does not include `.`

client/utils/needs-verify.js
codesankalp: Routing and redirection is based on `needsVerify` function that's why I have to change this to take effect in the organization-wrapper component.

nemesisdesign:
Superseded by https://github.com/openwisp/openwisp-wifi-login-pages/pull/489.

PR #491, opened 2021-12-13
[feature] Added option to upload logs to sentry #490

### LINKED ISSUE #490: [feature] Allow collection of backend errors with sentry
We should follow the sentry docs: https://docs.sentry.io/platforms/node/guides/express/.

pandafy:
Closes #490


env.json
```diff
@@ -0,0 +1,16 @@
...
+    "levelsMap": {
+      "silly": "debug",
+      "verbose": "debug",
+      "info": "info",
+      "debug": "debug",
+      "warn": "warning",
+      "error": "error"
+    }
+  }
+}
```
nemesisdesign, line 16: I think this file shouldn't be added in the repo, we should add something like `env.sample.json` and recommend users to rename it to `env.json`, and `env.json` should be added to gitignore. Can we call this `sentry-env.json`?

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #493, opened 2021-12-14
[fix] Registration fix on mini-browsers which do not have cookies #487

### LINKED ISSUE #487: [bug] App fails on mini-browsers which do not have cookies

pandafy:
Closes #487

### DIFF
--- a/client/components/mobile-phone-change/mobile-phone-change.js
+++ b/client/components/mobile-phone-change/mobile-phone-change.js
@@ -18,7 +18,6 @@ import logError from "../../utils/log-error";
 import handleChange from "../../utils/handle-change";
 import submitOnEnter from "../../utils/submit-on-enter";
 import Contact from "../contact-box";
-import handleSession from "../../utils/session";
 import validateToken from "../../utils/validate-token";
 import getError from "../../utils/get-error";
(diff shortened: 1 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -8,6 +8,7 @@ on:
   pull_request:
     branches:
       - master
+      - issues/490-enable-sentry
```
nemesisdesign, line 11: I think we can remove this

nemesisdesign (commented):
Looks good :+1:

nemesisdesign (approved):
Thanks

PR #537, opened 2022-02-08
[change] Do no generate exception on 400 bad request #536

### LINKED ISSUE #536: [change] Server: do not generate exception on 400 bad request response
This exception
```
Error: Request failed with status code 400
  File "/opt/openwisp2/wifi-login-pages/node_modules/axios/lib/core/createError.js", line 16, col 15, in createError
    var error = new Error(message);
  File "/opt/openwisp2/wifi-login-pages/node_modules/axios/lib/core/settle.js", line 17, col 12, in settle
    reject(createError(
  File "/opt/openwisp2/wifi-login-pages/node_modules/axios/lib/adapters/http.js", line 293, col 11, in IncomingMessage.handleStreamEnd
    settle(resolve, reject, response);
  File "node:events", line 402, col 35, in IncomingMessage.emit
  File "node:domain", line 475, col 12, in IncomingMessage.emit
  File "node:internal/streams/readable", line 1343, col 12, in endReadableNT
  File "node:internal/process/task_queues", line 83, col 21, in processTicksAndRejections
``` 

Is generated when calling the obtaintoken view and the response is 400 bad request.

This creates a new issue in error logging system, which is not useful at all, because it's not an uncaught exception but it's a regular error case.

In these cases we must log this as an info event and log the reasons for which the request failed (eg: print the body in the log message), we have to avoid to generate an exception.

codesankalp:
Closes #536

### DIFF
--- a/server/utils/logger.js
+++ b/server/utils/logger.js
@@ -106,9 +106,14 @@ try {
 
 export const logResponseError = (error) => {
   if (error.response) {
-    const {status, data} = error.response;
+    const {status, data, config} = error.response;
     if (status >= 500) Logger.error(error.response);
-    else Logger.info(`Request failed with ${status}: ${JSON.stringify(data)}`);
+    else
+      Logger.info(
+        `Request to ${config.url} failed with ${status}:\n${JSON.stringify(
+          data,
+        )}`,
+      );
   } else Logger.error(error);
 };
 
--- a/server/utils/logger.js
+++ b/server/utils/logger.js
@@ -107,7 +107,7 @@ try {
 export const logResponseError = (error) => {
   if (error.response) {
     const {status, data} = error.response;
-    if (status === 500) Logger.error(error.response);
+    if (status >= 500) Logger.error(error.response);
     else Logger.info(`Request failed with ${status}: ${JSON.stringify(data)}`);
   } else Logger.error(error);
 };
--- a/server/utils/logger.js
+++ b/server/utils/logger.js
@@ -103,7 +103,13 @@ try {
 } catch (error) {
   // no op
 }
-/* eslint-enable global-require */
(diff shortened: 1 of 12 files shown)


server/utils/logger.js
```diff
@@ -103,7 +103,13 @@ try {
 } catch (error) {
   // no op
 }
-/* eslint-enable global-require */
-/* eslint-enable import/no-unresolved */
+
+export const logResponseError = (error) => {
+  if (error.response) {
+    const {status, data} = error.response;
+    if (status === 500) Logger.error(error.response);
```
nemesisdesign, line 110: any 5xx error we should log with error level

nemesisdesign (changes requested):

server/utils/logger.js
codesankalp: Resolved in c8e7a80

server/utils/logger.js
```diff
@@ -103,7 +103,13 @@ try {
...
   // no op
 }
-/* eslint-enable global-require */
-/* eslint-enable import/no-unresolved */
+
+export const logResponseError = (error) => {
+  if (error.response) {
+    const {status, data} = error.response;
+    if (status >= 500) Logger.error(error.response);
+    else Logger.info(`Request failed with ${status}: ${JSON.stringify(data)}`);
```
pandafy, line 111: I think we should also add URL of the reponse if available. Otherwise, we will be clueless what endpoint generated the error. Is this information already included in data?

nemesisdesign: @codesankalp what about:
```
else Logger.info(`Request to ${url} failed with ${status}:\n${JSON.stringify(data)}`);
```

nemesisdesign (approved):

PR #539, opened 2022-02-10
[change] Removed Logger.error from controller files #540 

codesankalp:


codesankalp:
https://github.com/openwisp/openwisp-wifi-login-pages/blob/7ba8463c2e4fe6359b72eb4b549587566fdca4db/server/controllers/mobile-phone-token-controller.js#L102

These line will be called if response.status is undefined, should I remove it also?

codesankalp:
https://github.com/openwisp/openwisp-wifi-login-pages/pull/541

PR #541, opened 2022-02-10
[change] Removed Logger.error from controller files #540 

### LINKED ISSUE #540: [change] Remove Logger.error from controllers files
This error is not useful, we should remove it.

```
TypeError: Cannot read properties of undefined (reading 'status')
    at /opt/openwisp2/wifi-login-pages/server/controllers/user-radius-sessions-controller.js:47:38
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

It happens due to Logger.error in catch block of all the controller files.

codesankalp:
Closes #540


codesankalp:
image

nemesisdesign (approved):

PR #591, opened 2022-06-08
[fix] Show only email label in default org login form #589

### LINKED ISSUE #589: [bug] The default org should not show login by username or phone number
~~Currently It is not possible to login by username and phone number with the default org because the input field pattern limits it only for email.~~

~~**Default configuration for login form**:~~

~~https://github.com/openwisp/openwisp-wifi-login-pages/blob/3a179d55779ad98d5fced4ffda49ef5eea51af18/organizations/default/default.yml#L163-L166~~

~~See [username field configuration](https://github.com/openwisp/openwisp-wifi-login-pages#username-field-in-login-form) in README to enable login using username and phone number.~~

As the user gets registered using email so the username by default is email so we should show only email as the label for input username.

codesankalp:
Closes #589

### DIFF
--- a/client/components/login/__snapshots__/login.test.js.snap
+++ b/client/components/login/__snapshots__/login.test.js.snap
@@ -397,7 +397,7 @@ exports[`<Login /> interactions should not show phone_number field if auto_switc
             <label
               for="username"
             >
-              Email, phone number or username
+              Email
             </label>
             <input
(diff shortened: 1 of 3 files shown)


nemesisdesign (approved):

nemesisdesign:
Thanks! :+1:

PR #606, opened 2022-07-16
[Fix] Fixed captive portal login after registration #605

### LINKED ISSUE #605: [bug] Captive portal login after registration is not working
This is happening on the demo system and I verified this in my local dev env, I think it's happening only on master, because it was working fine in the last stable release.

See the GIF below, the captive portal login mock is not called and the session list is empty, while when it does work properly, the captive portal login mock creates the session and the app shows it.

cp-login-after-reg

**It would be great to have a failing test for this because it's a critical feature.**

codesankalp:
Fixes #605

### DIFF
--- a/browser-test/mobile-verfication.test.js
+++ b/browser-test/mobile-verfication.test.js
@@ -82,5 +82,10 @@ describe("Selenium tests for <MobileVerification />", () => {
       "div > p:nth-child(6) > span",
     );
     expect(await phoneElement.getText()).toEqual(data.phoneNumber);
+    const activeSessionTr = await getElementByCss(
+      driver,
+      "table tr.active-session",
+    );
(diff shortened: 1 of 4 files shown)


nemesisdesign (commented):
Maybe I understand now what's going on.

Probably this feature is broken only for simple email registration, but working for the mobile phone verification flow and the payment flow, because the other two flows redirect to other components before logging in the user.

Do we have a similar test also for mobile verification?
I mean, a test in which we register a user using mobile phone.

codesankalp:
Yes, it was only broken for email registration.
We have unit test for each component. But I think this test more requires complete flow.
Test is present for login. I added this in browser registration test here.
I will add it for mobile verification flow also.

nemesisdesign (approved):
Thanks :+1:

PR #665, opened 2023-06-24
[change] Do not send new SMS unless needed #656

### LINKED ISSUE #656: [change] Do not send new SMS unless needed
On some phones, clicking on an SMS to read the verification code will cause the mini-browser to close.
When the user opens the mini-browser again and gets logged in, the verification flow starts again and sends a new SMS, the old one not being valid anymore, this effectively causes the user to not be able to complete the SMS verification from their phone.

We shall modify OWLP and OpenWISP RADIUS to allow finding out if the user already has any valid Phone Token available, if so, we must not send a new code but allow them to use the previous one, they can send a new code if they want using the "resend" button.

This will need us to add another API endpoint which will just return either true or false, true if the user has a valid code, false if they don't. This will require changing OpenWISP RADIUS. We cannot rely on localStorage for this because it's not 100% reliable (some mini-browsers have limited support).

pandafy:
Closes #656

### DIFF
--- a/client/components/mobile-phone-verification/mobile-phone-verification.js
+++ b/client/components/mobile-phone-verification/mobile-phone-verification.js
@@ -176,7 +176,6 @@ export default class MobilePhoneVerification extends React.Component {
 
   async activePhoneToken() {
     const {orgSlug, language, userData} = this.props;
-    const {errors} = this.state;
     const url = mobilePhoneTokenStatusUrl(orgSlug);
     return axios({
       method: "get",
@@ -189,15 +188,20 @@ export default class MobilePhoneVerification extends React.Component {
     })
       .then((data) => data.active)
       .catch((error) => {
+        if (
+          error.response &&
+          error.response.status === 404 &&
+          error.response.data &&
+          error.response.data.response_code !== "INVALID_ORGANIZATION"
+        ) {
+          // This is kept for backward compatibility with older versions of OpenWISP RADIUS
+          // that does not have API endpoint for checking phone token status.
+          return false;
+        }
         const errorText = getErrorText(error);
         logError(error, errorText);
         toast.error(errorText);
-        this.setState({
-          errors: {
-            ...errors,
-            ...(errorText ? {nonField: errorText} : {nonField: ""}),
-          },
-        });
+        return errorText;
       });
   }
 
--- a/client/components/mobile-phone-verification/mobile-phone-verification.js
+++ b/client/components/mobile-phone-verification/mobile-phone-verification.js
@@ -14,6 +14,7 @@ import LoadingContext from "../../utils/loading-context";
 import {
   createMobilePhoneTokenUrl,
   verifyMobilePhoneTokenUrl,
+  mobilePhoneTokenStatusUrl,
 } from "../../constants";
 import getErrorText from "../../utils/get-error-text";
 import logError from "../../utils/log-error";
@@ -70,7 +71,9 @@ export default class MobilePhoneVerification extends React.Component {
       this.setState({phone_number});
       // send token via SMS only if user needs to verify
       if (!is_verified && settings.mobile_phone_verification) {
-        await this.createPhoneToken();
+        if (!(await this.activePhoneToken())) {
+          await this.createPhoneToken();
+        }
       }
     }
     setLoading(false);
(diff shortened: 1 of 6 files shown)


nemesifier (changes requested):
Looks good and works as expected during testing!
I have 2 requests.

## 1. Backward Compatibility

I think that if we try to use this version of OWLP with an older version of OpenWISP Radius, the SMS token verification process will break because the active token view is not available.

Can you please make the code resilient to this failure? Here's how it could work:

- if the nodeJS backend gets 404 from openwisp-radius, it shall forward the 404
- if the frontend JS code gets 404, it assumes it has to send a new verification code (old behavior is kept for backward compatibility).

What do you think?

## 2. I found a React Warning during testing

Can you see if you get this warning when reloading the phone number verification page?

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in Status (created by Connect(Status))
    in Connect(Status) (created by OrganizationWrapper)
    in Suspense (created by OrganizationWrapper)
```

If you do, can you look into whether this was there since before or not? If not please fix it, if it was already there please create an issue.

server/controllers/mobile-phone-token-controller.js
```diff
@@ -61,6 +61,57 @@ export const createMobilePhoneToken = (req, res) => {
...
+            });
+          }
+        });
+    }
+    return org.slug === reqOrg;
+  });
+  // return 404 for invalid organization slug or org not listed in config
+  if (!validSlug) {
+    res.status(404).type("application/json").send({
+      response_code: "INTERNAL_SERVER_ERROR",
```
nemesifier, line 110: Probably we don't want to send "INTERNAL_SERVER_ERROR" here but something different.
If this case is triggered, we must make sure the old behavior of sending the SMS token anyway is not triggered in this case because we know it's not needed.

client/components/mobile-phone-verification/mobile-phone-verification.js
```diff
@@ -171,6 +174,39 @@ export default class MobilePhoneVerification extends React.Component {
...
+      },
+      url,
+    })
+      .then((data) => data.active)
+      .catch((error) => {
+        if (error.response && error.response.status === 404) {
+          // This is kept for backward compatibility with older versions of OpenWISP RADIUS
+          // that does not have API endpoint for checking phone token status.
+          return false;
+        }
```
pandafy, line 196: The backend was already returning 404 response when the API endpoint is not present in OpenWISP RADIUS. I added this handling in the frontend.

server/controllers/mobile-phone-token-controller.js
```diff
@@ -61,6 +61,60 @@ export const createMobilePhoneToken = (req, res) => {
...
+    return org.slug === reqOrg;
+  });
+  // Return 401 for invalid organization slug or org not listed in config.
+  // This is different from other implementation because the frontend
+  // expects 404 only when the phone token status API endpoint is
+  // not implemented in OpenWISP RADIUS.
+  if (!validSlug) {
+    res.status(401).type("application/json").send({
+      response_code: "UNAUTHORIZED",
+    });
```
pandafy, line 114: Instead of using 404, I used 401 which makes it inconsistent with other views. 

When doing manual testing, if you visit http://localhost:8080/default/mobile-phone-verification for a random organization, it would show 404 page before this code is even executed.

pandafy: I have updated this code again.

server/controllers/mobile-phone-token-controller.js
```diff
@@ -61,6 +61,64 @@ export const createMobilePhoneToken = (req, res) => {
...
+          } catch (err) {
+            res.status(500).type("application/json").send({
+              response_code: "INTERNAL_SERVER_ERROR",
+            });
+          }
+        });
+    }
+    return org.slug === reqOrg;
+  });
+  // Return 401 for invalid organization slug or org not listed in config.
```
nemesifier, line 107: Is this comment right? Is 401 used here?

pandafy: Earlier, I used 401 here. But, that would not be correct. So, I updated the code to return 404 and used a different response code.

nemesifier (approved):

PR #701, opened 2023-09-10
Update mobile-phone-number-change-controller.js

harshsrivastava05:
Adding post method directly after axios instead of defining it inside it.


nemesifier (commented):
I don't understand the point of this change? Is anything broken or does anything need improvement?

harshsrivastava05:
it just looks better this way.

PR #702, opened 2023-09-10
Update mobile-phone-token-controller.js

harshsrivastava05:
Added post right next to axios instead of inside it.


nemesifier (commented):
Same as https://github.com/openwisp/openwisp-wifi-login-pages/pull/701#pullrequestreview-1655024347

PR #724, opened 2023-11-06
[feature] Added support for password expiration feature #713

### LINKED ISSUE #713: [feature] Add support for password expiration feature
Depends on https://github.com/openwisp/openwisp-radius/issues/491.

Modify login flow so that if validate auth token or obtain auth token return the new response format mentioned [in the previous point](https://github.com/openwisp/openwisp-radius/issues/491) the user will be taken to change the password
with an additional toast message which communicates that the password is expired and needs to be changed,

Ensure that once the password is changed the user is taken back to standard flow.

This needs a selenium browser test.

pandafy:
Closes #713

**Blockers**
- [x] https://github.com/openwisp/openwisp-users/pull/363
- [x] https://github.com/openwisp/openwisp-radius/pull/494


i18n/en.po
```diff
@@ -517,6 +517,10 @@ msgstr "Login successful"
 msgid "LOGOUT_SUCCESS"
 msgstr "Logout successful"
 
+#: client/components/status/status.js:119
+msgid "PASSWORD_EXPIRED"
+msgstr "Your password has expired, please change your password."
```
nemesifier, line 522:
```suggestion
msgstr "Your password has expired, please update it."
```

Please port the changes to the other languages using google translate.

nemesifier (changes requested):
Looks good, please see my comment below.

nemesifier (commented):
This should be ready, waiting for the PRs depended upon to be merged.

nemesifier (approved):

PR #744, opened 2024-01-08
[fix] FIxed package.json to ensure setup script is executed on build …

pandafy:
…and client


pandafy:
@nemesifier without this, the `yarn build` command was not executing `npm run setup`.

nemesifier (approved):

PR #858, opened 2025-03-29
[fix] Don't show "limit-info" div if API does not return any check #819

### LINKED ISSUE #819: [bug] Don't show "limit-info" div if API does not return any check
**Describe the bug**

If the logged-in user's RadiusGroup does not define any checks, then an empty box is shown where the user's usage is shown.

**Steps To Reproduce**

1. Setup WiFi Login Pages with OpenWISP RADIUS
2. Register with a user on WLP 
3. In OpenWISP, change the Radius Group of the user to `power-users`. 
4. Login to WLP with this user

**Expected behavior**
The empty div should not appear if the API endpoint returns an empty list for checks.

Add an automated test for this case.

**Screenshots**
Screenshot from 2024-09-11 22-41-15

pandafy:
## Checklist

- [ ] I have manually tested the changes proposed in this pull request.

Closes #819.

### DIFF
--- a/client/components/status/status.test.js
+++ b/client/components/status/status.test.js
@@ -1674,103 +1674,116 @@ describe("<Status /> interactions", () => {
   it("test getUserRadiusUsage method", async () => {
     jest.spyOn(toast, "error");
     jest.spyOn(toast, "dismiss");
-    axios
-      .mockImplementationOnce(() =>
-        Promise.reject({
-          response: {
-            status: 500,
-            headers: {},
-          },
-        }),
-      )
-      .mockImplementationOnce(() =>
-        Promise.resolve({
-          status: 200,
-          statusText: "OK",
-          data: {
-            checks: [
-              {
-                attribute: "Max-Daily-Session",
-                op: ":=",
-                value: "10800",
-                result: 0,
-                type: "seconds",
-              },
-            ],
-          },
-          headers: {},
-        }),
-      )
-      .mockImplementationOnce(() =>
-        Promise.resolve({
-          status: 200,
-          statusText: "OK",
-          data: {
-            plan: {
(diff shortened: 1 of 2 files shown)


client/components/status/status.test.js
```diff
@@ -1674,103 +1674,116 @@ describe("<Status /> interactions", () => {
...
+        response: {
+          status: 500,
+          headers: {},
+        },
+      }),
+    );
     wrapper.instance().getUserRadiusUsage();
     await tick();
     expect(wrapper.instance().state.radiusUsageSpinner).toBe(true);
+    expect(setTimeout).toHaveBeenCalledWith(expect.any(Function), 10000);
```
nemesifier, line 1697: What does this test line mean? A comment reminding us would be useful. I don't understand this.

nemesifier (commented):
I was testing this in my local env and the YAML didn't have `radius_usage_enabled: true`, so the div was not appearing at all. Is this intentional? 

It seems that it's disabled by default. Do you confirm?
We may want to change this in https://github.com/openwisp/openwisp-wifi-login-pages/issues/863.

The rest looks good, I left a comment below for a minor thing.

client/components/status/status.test.js
pandafy: In https://github.com/openwisp/openwisp-wifi-login-pages/pull/858/commits/7a528000f9e1033a9a56b89255917331ef4043cf, I have improved the code and test for better readability.

pandafy:
> I was testing this in my local env and the YAML didn't have `radius_usage_enabled: true`, so the div was not appearing at all. Is this intentional?

Yes, it is disabled by default. But, it is enabled for the `default` organization present for testing. 

https://github.com/openwisp/openwisp-wifi-login-pages/blob/09037d567c132aaefbc0b754218927cc7caa05d1/organizations/default/default.yml#L143-L148

Issue #808, opened 2024-08-04
[bug/tests] Add automated tests for "add-org" command 


nemesifier:
**Describe the bug**
I just updated a dependency and then figured out it may had been a bad idea: the dependency deals with interactive commands.
I figured out we may not have tests for this functionality, so I did a manual test to find out I was right and the feature indeed broke after the upgrade.
So I reverted it: https://github.com/openwisp/openwisp-wifi-login-pages/commit/c7d19111b70cb8d9b7fc7820e22aceaa9fda9fae.

We need to add a few basic tests for `yarn add-org` in the automated test suite.

**Steps To Reproduce**
Update inquirer to latest version, try using `yarn add-org`.

**Expected behavior**
The command should work, but it raises an exceptioon.

**System Informatioon:**

- OS: not relevant
- Node.js Version: Node.js 20
- Browser and Browser Version (if applicable): not relevant

PR #871, opened 2025-04-24
[test] Added test for add-org command #808

pandafy:
## Checklist

Related to #808

### DIFF
--- a/config/__tests__/add-org.test.js
+++ b/config/__tests__/add-org.test.js
@@ -1,0 +1,94 @@
+/* eslint-disable prefer-promise-reject-errors */
+const fs = require("fs");
+const path = require("path");
+const {spawnSync} = require("child_process");
+const yaml = require("js-yaml");
+
+// Test data and helpers
+const configDir = path.resolve(__dirname, "../../organizations");
+const testOrgSlug = "testorg";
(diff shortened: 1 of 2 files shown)


nemesifier (changes requested):
Why `config/__tests__/add-org.test.js`? It's pretty different from anything else we have here.
Would just `config/add-org.test.js` work?

I tried commenting out a critical line like:

```diff
diff --git a/config/add-org.js b/config/add-org.js
index edfa4d1..f13c9e4 100644
--- a/config/add-org.js
+++ b/config/add-org.js
@@ -219,7 +219,7 @@ const createConfiguration = async (response) => {
 
 const createConfigurationWithPrompts = async () => {
   const response = await prompt(prompts);
-  createConfiguration(response);
+  // createConfiguration(response);
 };
 
 const createConfigurationWithoutPrompts = (passedData) => {
```

The tests passed anyway. Is this correct? Doesn't look right to me, I would expect at least one test to fail.
The last test method seems to test prompts but doesn't check the result, I would at least check that given a series of prompts the result matches what we expect from the prompts.

nemesifier (approved):
As discussed, I'll merge for now but we'll need to complete the upgrade of the dependencies to add a full test.

PR #872, opened 2025-04-28
[feature] Allowed synchronous captive portal authentication #864

### LINKED ISSUE #864: [feature] Allow sending captive portal login and logout requests synchronously
**Is your feature request related to a problem? Please describe.**
On some mobile phones, the captive portal login or logout requests are not detected and the connectivity status is not changed when a user logs in or out.

**Describe the solution you'd like**
We can add a flag which allows to change the current behavior, which is to send the captive portal login and logout requests using an iframe, hence achieving captive portal login and out without reloading the page to a more standard behavior, which is to send the request synchronously (ie: normally, without the iframe) which will cause the browser to reload the page entirely and will be detected on all systems.

pandafy:
## Checklist

Closes #864

### DIFF
--- a/client/actions/logout.js
+++ b/client/actions/logout.js
@@ -8,6 +8,8 @@ const logout = (cookies, orgSlug, userAutoLogin = false) => {
     cookies.remove(`${orgSlug}_macaddr`, {path: "/"});
     sessionStorage.clear();
   }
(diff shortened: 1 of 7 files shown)


nemesifier (commented):
This looks good, I'll do some more testing before merging.

Issue #909, opened 2025-08-08
[change] Make progress bar use standard GB calculation


nemesifier:
The progress bar seems to show GiB, MiB, KiB, etc, so 1 KB equals 1024 bytes.

But what most users want is that 1 MB equals 1000000 bytes.

Let's change this.

PR #910, opened 2025-08-08
[fix] Display RADIUS usage in MB/GB (base 10) instead of MiB/GiB (base 2)

pandafy:
## Checklist

## Description of Changes

Previously, the status page used base 2 units when converting bytes to megabytes and gigabytes. While the units shown in the UI were labeled as MB/GB, the values were actually in MiB/GiB.

This patch updates the conversion logic to use base 10, so the displayed values now correctly match MB/GB.

Closes https://github.com/openwisp/openwisp-wifi-login-pages/issues/909


client/components/payment-process/payment-process.js
```diff
@@ -110,7 +110,8 @@ export default class PaymentProcess extends React.Component {
     }
 
     if (isTokenValid === true && !settings.payment_iframe) {
-      return this.redirectToPaymentUrl(userData.payment_url);
+      this.redirectToPaymentUrl(userData.payment_url);
+      return null;
```
pandafy, line 114: We were getting the following error on this component which redirects to the payment page 

```
Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

client/components/status/status.js
```diff
@@ -854,17 +854,15 @@ export default class Status extends React.Component {
         </td>
         <td>{this.getDuration(session.session_time)}</td>
         <td>
-          {bytes(session.output_octets, {
-            decimalPlaces: 0,
-            unitSeparator: " ",
-            unit: "MB",
+          {prettyBytes(session.output_octets, {
+            maximumFractionDigits: 0,
+            space: true,
```
pandafy, line 859: **ATTENTION**: Previously, we were forcing the conversion to show the RADIUS usage octets in MB. The new `pretty-bytes` library does not allow to specify the conversion and chooses the unit adaptively (e.g. for 1000 bytes, it will render the value as 1kB)

package.json
```diff
@@ -30,7 +29,7 @@
     "morgan": "^1.10.0",
     "node-plop": "^0.32.0",
     "nodemon": "^3.0.3",
-    "pretty-bytes": "^6.1.1",
+    "pretty-bytes": "5.6.0",
```
pandafy, line 32: The latest version of pretty-bytes have switched to ESM module. Thus, I had to downgrade it here to allow jest to run without problems.

package.json
```diff
@@ -70,6 +69,7 @@
     "@babel/preset-env": "^7.10.4",
     "@babel/preset-react": "^7.10.1",
     "@babel/runtime": "^7.12.5",
+    "@wojtekmaj/enzyme-adapter-react-17": "^0.8.0",
```
pandafy, line 72: yarn automatically moved this line to the top.

PR #915, opened 2025-08-26
[feature] Accounting swap octets on status page #900

### LINKED ISSUE #900: [feature/change] Accounting swap octets
In the status page, accounting octets are swapped by default (download is shown as upload and upload is shown as download).

We shall reverse this and allow to configure this with a new config option in the `status_page` section, we can call this option `accounting_swap_octets`, which should default to `false`.

https://github.com/openwisp/openwisp-wifi-login-pages/blob/2fb437e7b184ce21915353d0bfe511dcb7b873a5/client/components/status/status.js#L847-L860

https://github.com/openwisp/openwisp-wifi-login-pages/blob/2fb437e7b184ce21915353d0bfe511dcb7b873a5/client/components/status/status.js#L920-L945

Dhanus3133:
Closes #900

### DIFF
--- a/client/components/status/status.js
+++ b/client/components/status/status.js
@@ -109,6 +109,7 @@ export default class Status extends React.Component {
         );
 
         window.addEventListener("message", this.handlePostMessage);
+        window.postMessage({type: "owlp-ready"}, window.location.origin);
 
         if (macaddr) {
           cookies.set(`${orgSlug}_macaddr`, macaddr, {path: "/"});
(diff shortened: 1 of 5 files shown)


client/components/status/status.js
```diff
@@ -835,7 +835,15 @@ export default class Status extends React.Component {
   };
 
   getLargeTableRow = (session, sessionSettings, showLogoutButton = false) => {
-    const {language} = this.props;
+    const {language, statusPage} = this.props;
+    const {accounting_swap_octets} = statusPage;
+    let downloadOctets = session.output_octets;
```
nemesifier:
```suggestion
    let downloadOctets = session.input_octets;
```

client/components/status/status.js
```diff
@@ -835,7 +835,15 @@ export default class Status extends React.Component {
...
+    let uploadOctets = session.input_octets;
```
nemesifier:
```suggestion
    let uploadOctets = session.output_octets;
```

client/components/status/status.js
```diff
@@ -835,7 +835,15 @@ export default class Status extends React.Component {
...
+
+    if (accounting_swap_octets) {
+      downloadOctets = session.input_octets;
+      uploadOctets = session.output_octets;
```
nemesifier: invert

nemesifier (changes requested):
I double checked this and I verified that the initial implementation swapped octets by default:

photo_5879802252925980393_y

Upload is higher than download, but it's wrong.

We just need to invert these, see my comments below.

client/components/status/status.js
Dhanus3133: Updated.

client/components/status/status.js
Dhanus3133: Updated.

client/components/status/status.js
Dhanus3133: Updated.

PR #935, opened 2025-09-25
[fix] Use calling_station_id instead of macaddr for MAC filtering #914

### LINKED ISSUE #914: Status component uses incorrect query param for session lookup
**Describe the bug**
The Status component in the client currently checks for an active session by filtering with the `macaddr` query parameter. However, the OpenWISP RADIUS API endpoint does not support filtering by `macaddr`; it expects `calling_station_id` as the query param for MAC address-based filtering.

**Relevant code:**
-[ `client/components/status/status.js` (uses `macaddr` as query param)](https://github.com/openwisp/openwisp-wifi-login-pages/blob/e1ef7e0f7ad910faf51ff04c8ec878a865835e90/client/components/status/status.js#L217-L231)
- [The RADIUS API (`openwisp_radius/api/views.py`) expects `calling_station_id` for MAC address filtering.](https://github.com/openwisp/openwisp-radius/blob/4d9d85195624b60f157339e8884ac45539918f30/openwisp_radius/api/views.py#L808-L848)

**Steps To Reproduce**
1. Attempt to retrieve radius sessions for a user with a specific MAC address.
2. Observe that the frontend sends a `macaddr` query param to the API.
3. The API does not filter sessions as expected because it does not recognize `macaddr`.

**Expected behavior**
The frontend should use the `calling_station_id` query parameter when requesting sessions by MAC address, matching the OpenWISP RADIUS API's expectations.

pranshustuff:
## Checklist

Closes #914 

## Description of Changes

Status component used param `macaddr` to filter
users by MAC address but Radius session uses the
param `calling_station_id`. Caused it to return all sessions instead of filtering properly.

Changed params for getUserActiveRadiusSessions(params) from `{ macaddr }` to `{ calling_station_id: macaddr }`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,11 @@ jobs:
           key: ${{ runner.os }}-pip-${{ env.OW_RADIUS_VERSION }}
           restore-keys: ${{ runner.os }}-pip-
 
+      - name: Start InfluxDB and Redis container
+        run: |
+          cd openwisp-radius
+          docker compose up -d redis
+
       - name: Installing OpenWISP Radius
         run: |
           cd openwisp-radius


client/components/status/status.js
```diff
@@ -218,7 +218,7 @@ export default class Status extends React.Component {
 
       const macaddr = cookies.get(`${orgSlug}_macaddr`);
       if (macaddr) {
-        const params = {macaddr};
+        const params = {calling_station_id: macaddr};
```
nemesifier, line 221: There's at least one more occurrence of `const params = {macaddr}` in `handleLogout`.

nemesifier (changes requested):
Look at status.test.js, we have several tests in which we check that `getUserActiveRadiusSessions`. We can add one test which verifies it gets called with the right param.

client/components/status/status.js
pranshustuff: Fixed!

Issue #918, opened 2025-09-08
[change] Refactor status component to simplify logic and improve maintainability


pandafy:
The current status component in `client/components/status/status.js` has grown to be very large and complex. It handles multiple responsibilities, including authentication, verification, payment, session management, captive portal logic, and UI rendering. This makes the codebase difficult to understand and maintain. The complexity also increases the risk of bugs and makes testing more difficult.

I opened this issue to gather information. **We need to discuss the changes before going ahead with this issue**

PR #948, opened 2025-10-08
[fix] Removed check for active session for performing captive portal login

pandafy:
## Checklist

## Description of Changes

The status component will perform captive portal login when mustLogin is set to true. 

If the user has registered with "bank_card" method, and payment requires internnet, then it will perform captive portal login when mustLogin is set to true.

The status component will not check if the user is already logged in based on the mac address reported by the captive portal. 

Related to https://github.com/openwisp/openwisp-wifi-login-pages/issues/918.

### DIFF
--- a/client/components/status/status.test.js
+++ b/client/components/status/status.test.js
@@ -357,28 +357,6 @@ describe("<Status /> interactions", () => {
     });
   });
 
-  it("should call getUserActiveRadiusSessions with calling_station_id", async () => {
-    axios.mockImplementationOnce(() => Promise.resolve({data: []}));
-    props = createTestProps({
-      location: {
-        search: "?macaddr=4e:ed:11:2b:17:ae",
-      },
-      userData: {...responseData, mustLogin: true},
-    });
-    validateToken.mockReturnValue(true);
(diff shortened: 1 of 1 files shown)


client/components/status/status.js
```diff
@@ -216,53 +216,16 @@ export default class Status extends React.Component {
...
-            this.notifyCpLogin(userData);
-            this.loginFormRef.current.submit();
-          }
-        }
-        if (captivePortalSyncAuth) {
-          this.finalOperations();
-        }
-      } else if (this.loginFormRef && this.loginFormRef.current && mustLogin) {
-        if (
-          method === "bank_card" &&
```
nemesifier, line 246: what about this part?

.github/workflows/ci.yml
```diff
@@ -129,6 +129,7 @@ jobs:
           cp browser-test/local_settings.py \
              openwisp-radius/tests/openwisp2/local_settings.py
           cd openwisp-radius
+          docker compose up -d redis
```
nemesifier: Let's remove this as it's now handled in master already.

PR #955, opened 2025-10-30
[feature] Logout user from WLP if RADIUS session is terminated #933

### LINKED ISSUE #933: [feature] Status page may say we're logged in but RADIUS session may not be active
**Description** 
 
When leaving the WiFi Login Pages status page open in the browser on mobile or desktop, the page continues showing *“logged in”* even after a RADIUS session has expired or connectivity is lost. This causes confusion because the UI indicates there's an active session, while other apps report no internet connection.

**Steps to Reproduce**
1. Connect to WiFi and log in
3. Status page shows *“logged in”*.
4. Leave device inactive for the time needed to trigger Idle-Timeout
5. Return to Chrome → page still shows *“logged in”*.
6. Switch to another app → no internet connection.

**Expected Behavior**  
When the user returns to the login page (from inactivity or another tab), the status should automatically refresh and reflect the actual session state (e.g. *“logged out”* if the session expired).

**Implementation notes**
I think we should use a combination of the following:

- Detect the mac address of the user thanks to the querystring param sent by the captive portal, we should already have logic for this, let's check it still works
- When we get the list of RADIUS sessions, if there's no open session for the current mac address and we are not in "internet-mode", we should switch to this new "logged in but offline" mode (let's find a better name for this)

pandafy:
## Checklist

Closes #933

## Description of Changes

Ensure the user is logged out from WiFi Login Pages when their NAS session is no longer active (e.g. due to Idle-Timeout or disconnection). This keeps the UI state in sync with actual RADIUS sessions and avoids showing “logged in” after the session expires.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Change log
 
+## Version 1.3.0 [unreleased]
+
+Work in progress.
+
 ## Version 1.2.0 [2025-10-24]
 
 ### Features
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openwisp-wifi-login-pages",
-  "version": "1.2.0",
+  "version": "1.3.0-dev",
   "description": "",
   "main": "src/index.js",
   "dependencies": {


client/components/status/status.js
```diff
@@ -500,6 +501,35 @@ export default class Status extends React.Component {
     await this.getUserRadiusSessions(para);
   }
 
+  async logoutIfCurrentRadiusSessionIsInactive() {
+    // If OpenWISP no longer reports an active session for this MAC address,
+    // the NAS has likely terminated the connection (e.g. due to Idle-Timeout).
+    // In that case, log out the user from the WiFi Login Pages to keep the UI
+    // consistent and allow them to reconnect quickly.
+    const {internetMode, cookies, orgSlug} = this.props;
+
```
nemesifier:
```suggestion
```

client/components/status/status.js
```diff
@@ -500,6 +501,35 @@ export default class Status extends React.Component {
...
+    if (!mac) {
+      // Cannot determine session state without a MAC address.
+      return;
+    }
+
+    const {activeSessions} = this.state;
+    const isCurrentSessionActive = activeSessions.some(
+      (session) => session.calling_station_id === mac,
+    );
+
```
nemesifier:
```suggestion
```

client/components/status/status.js
```diff
@@ -500,6 +501,35 @@ export default class Status extends React.Component {
...
+    if (!isCurrentSessionActive) {
+      // Session for this MAC is no longer active on the NAS — log out from WLP.
+      // The "true" argument ensures the user can quickly sign in again.
```
nemesifier, line 526: Don't you think we should add a toast message here to inform the user?

nemesifier (approved):

PR #1006, opened 2025-12-29
[deps] Upgraded React to 18.3 #870

### LINKED ISSUE #870: [deps] Upgrade to latest version of react
## Recommended Upgrade Path

Upgrade to React 18.3 First: React 18.3 includes deprecation warnings for features removed in React 19, helping identify necessary code changes.​

## Run Codemods:

Utilize the official React 19 codemods to automate common refactorings, eg:

    npx codemod@latest react/19/migration-recipe

## Update Dependencies

Ensure all dependencies are compatible with React 19.

Check for updates or replacements for packages that are no longer maintained.​

Dependencies depending on react:
- react-dom: `{ react: '^19.1.0' }`
- react-cookie: `{ react: '>= 16.3.0' }`
- react-countdown: `{ react: '>= 15', 'react-dom': '>= 15' }`
- react-helmet: { react: '>=16.3.0' }
- react-side-effect: `{ react: '^16.3.0 || ^17.0.0 || ^18.0.0' }`
- react-infinite-scroll-component: `{ react: '>=16.0.0' }`
- react-phone-input: `{ lodash: '^4.11.1', react: '^0.14.8 || ^15.0.1' }`
- react-redux:
  ```
  {
    '@types/react': '^18.2.25 || ^19',
    react: '^18.0 || ^19',
    redux: '^5.0.0'
  }
  ```
- react-router-dom: `{ react: '>=18', 'react-dom': '>=18' }`
- react-router: `{ react: '>=18', 'react-dom': '>=18' }`
- react-select:
  ```
  {
    react: '^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0',
    'react-dom': '^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0'
  }
  ```
- react-transition-group: `{ react: '>=16.6.0', 'react-dom': '>=16.6.0' }`
- use-isomorphic-layout-effect: `{ react: '^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0' }`
- react-test-renderer: `{ react: '^19.1.0' }`
- react-toastify: `{ react: '^18 || ^19', 'react-dom': '^18 || ^19' }`

## Test Thoroughly

After making the necessary changes, run your test suite to catch any regressions or issues introduced during the upgrade.​

PRoIISHAAN:
## Checklist

Closes #870.

## Description of Changes

1. React 18 Migration
   Upgraded React and React DOM from v17 → v18.3
   Replaced legacy ReactDOM.render with createRoot
   Migrated from react-helmet → react-helmet-async for React 18 compatibility

2. Migration from Enzyme to RTL
    Full migration from Enzyme to React Testing Library (RTL)
    Added RTL + Jest DOM setup:
        @testing-library/react
        @testing-library/jest-dom

3. Test Suite Migration
    Updated snapshot tests to match RTL rendering
    Rewrote tests to use RTL patterns (fireEvent, screen, waitFor)
    Fixed async test stability issues caused by React 18 concurrent rendering
    All test suites now pass

4. Server-Side Refactor for Testability
    Extracted Express app into server/app.js
    Simplified server/index.js to only handle startup logic
    Updated API tests to use supertest instead of live HTTP requests

6. Build & Tooling Updates
    Updated Babel plugins for React 18 compatibility
    Updated Jest to v30
    Updated Webpack config:
        Added eslint-webpack-plugin
    Improved linting behavior (non-blocking)

Updated package.json


client/components/registration/registration.js
```diff
@@ -171,7 +170,7 @@ export default class Registration extends React.Component {
     let plan_pricing;
     if (selectedPlan !== null) {
       plan_pricing = plans[selectedPlan];
-      postData.plan_pricing = plan_pricing.id;
+      postData.planPricing = plan_pricing.id;
```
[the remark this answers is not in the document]
nemesifier: I think the AI is right on this one.

client/components/registration/registration.js
```diff
@@ -718,10 +730,11 @@ export default class Registration extends React.Component {
                               value={zipcode}
                               onChange={this.handleChange}
                               autoComplete="postal-code"
+                              aria-label="Zip Code"
                             />
                           </div>
                           <div className="row tax_number">
-                            <label htmlFor="tax_number">{t`TAX_NUMBER_LBL`}</label>
+                            <label htmlFor="tax_number">{t`taxNumber_LBL`}</label>
```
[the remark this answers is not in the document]
nemesifier: Same here

nemesifier (commented):
Please follow up with all the reviews from coderabbit, if you don't agree to something just reply to it.

nemesifier (commented):
Now that #1007 is merged, some conflicts appeared.

client/components/login/login.test.js
```diff
@@ -438,344 +529,436 @@ describe("<Login /> interactions", () => {
...
+      email: "tester@test.com",
+      phone_number: "+393660011333",
+      username: "+393660011333",
+      key: "test-token", // Use key instead of auth_token for API response
+      radius_user_token: "jwyVSZYOze16ej6cc1AW5cxhRjahesLzh1Tm2y0d",
+      first_name: "",
+      last_name: "",
+      birth_date: null,
+      location: "",
+    };
```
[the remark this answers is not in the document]
PRoIISHAAN: i dont think this is relevant to my pr

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "",
   "main": "src/index.js",
   "dependencies": {
+    "@babel/plugin-transform-private-methods": "^7.27.1",
```
nemesifier: isn't this a dev dependency?

package.json
```diff
@@ -4,7 +4,9 @@
...
     "@babel/register": "^7.10.1",
+    "@sentry/node": "^10.34.0",
```
nemesifier: why?

package.json
```diff
@@ -72,26 +76,26 @@
...
+    "eslint-plugin-jest-dom": "^5.5.0",
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-react": "^7.20.0",
     "eslint-plugin-react-hooks": "^4.2.0",
+    "eslint-plugin-testing-library": "^7.15.3",
+    "eslint-webpack-plugin": "^4.0.1",
     "html-webpack-plugin": "^4.5.2",
-    "jest": "^30.2.0",
+    "jest": "^29.7.0",
```
nemesifier: why is this being downgraded?

nemesifier (commented):
I haven't been able to  review  the full diff yet, please see my questions below.

package.json
PRoIISHAAN: i am sorry, i had it in dev dependencies as well, i have removed it from dependencies now.

package.json
PRoIISHAAN: I must have made an error while resolving merge conflicts, Fixed this.

package.json
PRoIISHAAN: earlier @sentry/node was automatically resolved and locked in yarn.lock through winston-transport-sentry-node's peer dependency.
During the React 18 migration, the yarn.lock file was regenerated due to the major dependency updates.
The winston-transport-sentry-node package lists @sentry/node as a peer dependency, adding it explicitly to package.json ensures it's always installed regardless of how the package manager resolves peer dependencies.

nemesifier (changes requested):
I am sorry I couldn't look into this sooner, there's been a number of urgent matters to handle before this.

The app works after the upgrade, so it's encouraging!

Can you please rebase your changes on the latest master and resolve conflicts?

See also my comments below and let me know what you think.

## Bugs / Correctness

**`delete data.key` removal changes `setUserData` payload silently**
(`client/components/login/login.js:305-306`)

```js
// Before: key removed from data, only auth_token present in spread
const {key: auth_token} = data;
delete data.key;
setUserData({...data, auth_token, mustLogin: true});

// After: both `key` and `auth_token` end up in the spread
const {key: auth_token} = data;
setUserData({...data, auth_token, mustLogin: true});
```

The `delete` was removed, so `setUserData` now receives an object with both `key` and
`auth_token`. Whether this matters depends on what reducers do with `key` downstream,
but it's a silent behavioral change that should be intentional or reverted.

**`link.url` used as React list key without uniqueness guarantee**
(`header.js:151`, `header.js:231`)

Using `key={link.url}` fixes the index-as-key anti-pattern but silently breaks rendering
if two links share the same URL. A composite key like `${link.url}-${link.text}` would
be safer.

## Dead Code

**`@sentry/node` added to production dependencies but never used**

`package.json` adds `"@sentry/node": "^7.1.1"` under `dependencies`, but there is no
`import`/`require` of it anywhere in `server/` or `client/`. Should be removed.

**`config/setupTests.js` is unreachable**

Duplicates the `matchMedia` mock and `@testing-library/jest-dom` import already in
`config/jest.config.js`, but is not referenced under `setupFilesAfterEnv` in
`package.json` so it never runs. Either wire it up and remove the duplication from
`jest.config.js`, or delete it.

**Orphaned JSDoc blocks in `client/test-utils.js` (lines 16–33)**

Two JSDoc blocks document `renderWithRouter` and `renderWithRedux` functions that were
never implemented. No code follows either comment. Should be removed.

## Misc Issues

**`getLoadingContextValue()` doesn't fix the context re-render problem**
(`organization-wrapper.js`)

The original code suppressed `react/jsx-no-constructed-context-values` with an
eslint-disable comment. The refactor moves object construction to a method, but
`this.getLoadingContextValue()` is still called inside `render()`, producing a new
object reference on every render — the same underlying problem. The disable comment was
the honest solution; this gives a false impression of a fix.

**`camelcase: "off"` disabled globally in ESLint** (`.eslintrc.json`)

Turns off the camelcase rule project-wide to avoid per-file `eslint-disable` comments in
files that destructure snake_case backend properties. A narrower fix would be
`"camelcase": ["error", {"allow": ["auth_token", "is_active", ...]}]`. The global
disable will silently permit new snake_case identifiers going forward.

**`eslint-webpack-plugin` configured to never fail** (`webpack.config.js`)

```js
new ESLintPlugin({ failOnError: false, failOnWarning: false })
```

Adds ESLint to every webpack build but never blocks on violations. It increases build
time without enforcement. Either remove it (ESLint already runs in CI) or set
`failOnError: true`.

**PR description says `status.test.js` is skipped but it isn't**

The description states the file is *"kept as .skip / backup for future migration"*. In
reality it has been fully migrated with no `describe.skip` or `test.skip`. The
description needs updating.

PRoIISHAAN:
@nemesifier (delete data.key removal changes setUserData payload silently
(client/components/login/login.js:305-306)) I did this because I was getting an eslint error (no-param-reassign), should I revert this and add an eslint disable?

nemesifier:
> @nemesifier (delete data.key removal changes setUserData payload silently
> (client/components/login/login.js:305-306)) I did this because I was getting an eslint error (no-param-reassign), should I revert this and add an eslint disable?

@PRoIISHAAN I'd let it fail for now so I can double check and decide what to do.
We should do weird code changes just to satisfy the linter if we're not confident that the changes are really needed.

PR #1015, opened 2026-01-21
[change:ux] Make progress bar in status page more user friendly #863

### LINKED ISSUE #863: [UX] Make progress bar in status page more user friendly
**The opinion from UX experts is needed before moving forward with this.**

I believe we should enhance the progress bar on the status page to better indicate the time and/or traffic available to the user.

Image

The main issue I see here is that the unit of measure for the remaining time is missing, I fear users may not easily understand what 03:00:00 means.

Maybe we can communicate the remaining time/traffic too?

PRoIISHAAN:
## Checklist

Closes #863 

## Description of Changes

Added react-circular-progressbar package for circular progress visualization
Replaced linear <progress> bars with circular progress indicators (CircularProgressbarWithChildren)
Added new CSS classes for the redesigned usage overview section
New Helper Methods in status.js:
getUserCheckFormattedValue() - Updated to calculate and format remaining quota
getUserCheckUsedValue() - New method to format "used of total" display
getResetTimeRemaining() - New method to calculate time until quota reset
getUsageColorAndIcons() - New method to determine color and icons based on usage percentage

## Screenshot

Screenshot 2026-01-21 at 2 39 54 PM


nemesifier:
@avocadodavinci if you could chime in here it would be very helpful!

nemesifier (commented):
@PRoIISHAAN we can use the horizontal bar for the desktop view and radial for narrow screens, see the [figma design](https://www.figma.com/design/0eaWAMSiEAgfHMIjIVRv2F/Open-WISP-Bridgebeam?node-id=1614-4722).

client/components/status/status.js
```diff
@@ -1169,18 +1170,220 @@ export default class Status extends React.Component {
...
+      return {
+        color: "#FBBF24",
+        timerIcon: "/assets/default/timerIconYellow.svg",
+        dataIcon: "/assets/default/dataIconYellow.svg",
+      };
+    }
+    return {
+      color: "#DB3B21",
+      timerIcon: "/assets/default/timerIconRed.svg",
+      dataIcon: "/assets/default/dataIconRed.svg",
```
nemesifier, line 1306: **Icons don't exist and the path scheme is wrong.**

None of the six referenced SVGs (`timerIcon{Green,Yellow,Red}.svg`, `dataIcon{Green,Yellow,Red}.svg`) exist anywhere in the repo: in production these `` requests will 404 and users will see broken-image placeholders. The tests only string-compare URLs so they don't catch it.

Also, `/assets/<slug>/` (see `config/setup.js:158-165`) is per-org branding (logos, social icons, custom CSS). Hardcoding `/assets/default/...` means every organization deploying WLP has to drop six SVGs into their own `client_assets/` or the icons 404 for them too.

Please add the six SVGs under `client/assets/`, alongside `warning.svg` / `eye.svg` / `eye-slash.svg`, import them here, and update the test/snapshot.

client/components/status/status.js
```diff
@@ -1169,18 +1170,220 @@ export default class Status extends React.Component {
...
+    return t`TIME_LESS_THAN_MINUTE`;
+  };
+
+  // eslint-disable-next-line class-methods-use-this
+  getUsageColorAndIcons = (value, result) => {
+    const numValue = Number(value);
+    const numResult = Number(result);
+
+    // Default to green (low usage) if value is 0 or non-numeric
+    if (!numValue || Number.isNaN(numValue)) {
```
nemesifier, line 1279: Dead branch: `!numValue` is already true for `NaN`, `0`, `undefined`, `null`, `""`, so the `Number.isNaN` arm never fires.

Clearer:
```js
if (!Number.isFinite(numValue) || numValue <= 0) {
```

client/components/status/status.js
```diff
@@ -1169,18 +1170,220 @@ export default class Status extends React.Component {
...
+    const remaining = Math.max(0, intValue - intResult);
     switch (type) {
       case "bytes":
-        return intValue === 0 ? 0 : filesize(intValue, {round: 2});
-      case "seconds":
-        return timeFromSeconds(intValue);
+        return remaining === 0 ? t`0` : `${filesize(remaining, {round: 2})}`;
+      case "seconds": {
+        if (remaining === 0) {
+          return t`0`;
```
nemesifier, line 1187: `` t`0` `` is meaningless: the digit zero is language-neutral. Return plain `"0"`.

client/components/status/status.js
```diff
@@ -1169,18 +1170,220 @@ export default class Status extends React.Component {
...
+        return `${hours}${t`TIME_HOUR_ABBR`} ${minutes}${t`TIME_MINUTE_ABBR`}`;
+      }
+      if (hours > 0) {
+        return `${hours}${t`TIME_HOUR_ABBR`}`;
+      }
+      if (minutes > 0) {
+        return `${minutes}${t`TIME_MINUTE_ABBR`}`;
+      }
+      return t`TIME_LESS_THAN_MINUTE`;
+    };
```
nemesifier, line 1231: This `formatTime` logic is duplicated three times: here (1218-1231), in `getUserCheckFormattedValue` (1191-1200), and in `getResetTimeRemaining` (1261-1270). Extract one helper method and call it from all three. The hour/minute concatenation pattern (`${hours}${t`TIME_HOUR_ABBR`} ${minutes}${t`TIME_MINUTE_ABBR`}`) also bakes English word order into every language: prefer ttag interpolation so translators control the order.

client/components/status/status.js
```diff
@@ -1169,18 +1170,220 @@ export default class Status extends React.Component {
...
+                check.value,
+                check.type,
+                check.result,
+              )}
+            </div>
+          </div>
+        </div>
+      </div>
+    );
+  };
```
nemesifier, line 1385: `renderUsageCheckContentSmall` and `renderUsageCheckContentBig` share the same outer shell: `usage-check-content` > `usage-check-header` with identical icon/label markup, and both pass `(check.value, check.type, check.result)` into the same two helpers. The progress widget and bottom layout legitimately differ, but the wrapper and header should be factored into a single helper that both variants compose, so changes don't need to be made twice.

client/components/status/status.js
```diff
@@ -1169,18 +1170,220 @@ export default class Status extends React.Component {
...
+          <div className="usage-progress-text-bottom">
+            <div style={{fontWeight: "400", color: "#71717A"}}>
+              {this.getUserCheckFormattedValue(
+                check.value,
+                check.type,
+                check.result,
+              )}{" "}
+              {t`USAGE_REMAINING`}
+            </div>
+            <div style={{color: "#52525B"}}>
```
nemesifier, line 1374: Inline styles with hardcoded hex colors. The rest of the component uses class names from `index.css`; please move these into CSS so the file stays consistent and theming stays easy.

client/components/status/status.js
```diff
@@ -1254,61 +1457,96 @@ export default class Status extends React.Component {
...
+                <div className="usage-overview-title">{t`DAILY_USAGE_OVERVIEW`}</div>
+                <p>{t`DAILY_USAGE_OVERVIEW_DESCRIPTION`}</p>
+                {radiusUsageSpinner ? this.getSpinner() : null}
+                {settings.subscriptions && userPlan.name && (
+                  <h3>
+                    {t`CURRENT_SUBSCRIPTION_TXT`} {userPlan.name}
+                  </h3>
+                )}
+                {userChecks && (
+                  <div style={{width: "100%"}}>
```
nemesifier, line 1470: Inline `style={{width: "100%"}}` — move to a CSS class in `index.css` for consistency with the rest of the component.

client/components/status/status.js
```diff
@@ -1254,61 +1457,96 @@ export default class Status extends React.Component {
...
+                          </React.Fragment>
+                        );
+                      })}
+                    </div>
+                    {userChecks[0]?.reset &&
+                      this.getResetTimeRemaining(userChecks[0].reset) && (
+                        <div className="usage-reset-info">
+                          {`*${t`DAILY_LIMITS_RESET_IN`} ${this.getResetTimeRemaining(userChecks[0].reset)}`}
+                        </div>
+                      )}
```
nemesifier, line 1528: A few issues bundled here:

**Only inspects `userChecks[0]`.** If the backend ever returns the `reset` field on a check other than the first (e.g. only on the seconds check, with the bytes check at index 0), the banner silently disappears. Safer:
```js
const resetTs = userChecks.find(c => c.reset)?.reset;
```

**`getResetTimeRemaining` runs twice per render** (lines 1524 and 1526). Hoist it into a `const` above the JSX.

**`DAILY_LIMITS_RESET_IN` is concatenated, not interpolated.** This locks every language into putting the time at the end. Use ttag interpolation so translators control word order:
```js
t`*Daily limits reset in ${resetInfo}`
```

**The leading `*` has no footnote.** It implies "see note below" but nothing follows it on the page. Either drop the asterisk or add the line it refers to.

nemesifier (changes requested):
Thanks @PRoIISHAAN, it looks on the right track! :clap: 
A few things still outstanding before this is ready to merge.

I'll be doing more manual testing in the coming days to stress test it further, in the meantime see my comments below.

**Run a translation extraction.** The `#:` references in the `.po` files are stale. For example `en.po:178-180` claims `PLAN_UPGRADE_BTN_TXT` is at `status.js:1193`, but that's now inside `getUserCheckFormattedValue`; the actual key sits at line 1546. Please re-run `update-translations.sh` (or its yarn equivalent) before merging.

PRoIISHAAN:
@nemesifier while running the yarn "translations-update" script, I am getting an Error: "You can not use Identifier 'message' as an argument to gettext" on line number 772 status.js

pandafy:
@coderabbitai full review!

nemesifier:
@coderabbitai full review!

client/components/status/status.js
```diff
@@ -1161,18 +1197,199 @@ export default class Status extends React.Component {
...
     }
+    const used = Math.min(resultNumber, total);
+    let usedFormatted = used;
+    let totalFormatted = total;
+    if (type === "bytes") {
+      usedFormatted = formatUsageBytes(used);
+      totalFormatted = formatUsageBytes(total);
+    } else if (type === "seconds") {
+      usedFormatted = formatUsageTime(used);
+      totalFormatted = formatUsageTime(total);
```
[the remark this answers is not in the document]
pandafy: @coderabbitai this is intentional

PR #1047, opened 2026-03-03
[change] Rebuild webpack when there is change in organizations dir #425

### LINKED ISSUE #425: [change] Rebuild webpack when their is change in organizations dir
Rebuild webpack whenever there is a change in organizations dir to avoid:
- rebuild each time when change is made
- edit the file which gets copied instead of the source

PRoIISHAAN:
## Checklist

Closes #425 .

## Description of Changes

Add OrganizationsWatchPlugin to webpack.config.js that watches all files under organizations/ during development. When any file there changes, setup.writeConfigurations() is re-run automatically before the next rebuild, so developers always edit the source files instead of the copied counterparts.
Also fix writeConfigurations() in setup.js to clear the module-level arrays at the start of each call, preventing duplicate entries when the function is invoked multiple times in a single session.


Viscous106 (approved):
by mistake

nemesifier (commented):
Sorry for my late response, I did some testing and although it works initially, it's not consistent nor reliable.

I tried changing several parameters of the default.yml org like the text in components.header.links, the main org name or footer text, it did work for the first change but then it stopped.

PRoIISHAAN:
@nemesifier I tried changing the parameters on my end, the ones mentioned by you and other parameters too, and it works as intended
[demonstration video](https://drive.google.com/file/d/1A_BJAY-UiAjNyWVEoTqr_XChdsTiD_1Z/view?usp=sharing)

PR #1061, opened 2026-03-15
[tests] Make add-org interactive test catch config creation regressions

### LINKED ISSUE #880: [test] Make add-org automated more robust: interactive input
Follow up of #808 and #871.

Why `config/__tests__/add-org.test.js`? It's pretty different from anything else we have here.
Would just `config/add-org.test.js` work?

I tried commenting out a critical line like:

```diff
diff --git a/config/add-org.js b/config/add-org.js
index edfa4d1..f13c9e4 100644
--- a/config/add-org.js
+++ b/config/add-org.js
@@ -219,7 +219,7 @@ const createConfiguration = async (response) => {
 
 const createConfigurationWithPrompts = async () => {
   const response = await prompt(prompts);
-  createConfiguration(response);
+  // createConfiguration(response);
 };
 
 const createConfigurationWithoutPrompts = (passedData) => {
```

The tests passed anyway. Is this correct? Doesn't look right to me, I would expect at least one test to fail.
The last test method seems to test prompts but doesn't check the result, I would at least check that given a series of prompts the result matches what we expect from the prompts.

czarflix:
## Checklist

Closes #880.

## Description of Changes

The existing interactive add-org test was only a smoke test, so commenting out
`createConfiguration(response)` in `createConfigurationWithPrompts()` still left
the focused test suite green.

This updates `config/__tests__/add-org.test.js` to add a regression test that:

- runs the interactive CLI path with mocked prompt answers
- checks that the organization config file is created
- verifies key YAML fields from the prompt data

The existing smoke test is kept and no application code is changed.

## Verification

- `yarn jest config/__tests__/add-org.test.js --runInBand`
- `./run-qa-checks`
- confirmed the new test fails if `createConfiguration(response)` is commented out
- `yarn test --runInBand` was compared against untouched `origin/master`; the observed login/header failures are preexisting and unrelated to this change

## Screenshot

N/A - test-only change

### DIFF
--- a/config/__tests__/add-org.test.js
+++ b/config/__tests__/add-org.test.js
@@ -26,6 +26,64 @@ const validData = {
   assets_confirm: false,
 };
 
+const interactiveData = {
+  ...validData,
+  name: "Interactive Test Organization",
+  uuid: "interactive-test-uuid",
+  secret_key: "interactive-test-secret",
+};
+
+const interactiveRunner = `
+  const fs = require("fs");
+  const path = require("path");
+  const response = JSON.parse(process.argv[1]);
+  const outputPath = path.resolve(
+    "organizations",
+    response.slug,
+    \`\${response.slug}.yml\`,
+  );
+
+  process.on("uncaughtException", (error) => {
+    console.error(error);
+    process.exit(1);
+  });
+
+  process.on("unhandledRejection", (error) => {
+    console.error(error);
+    process.exit(1);
+  });
+
+  const inquirerPath = require.resolve("inquirer");
+  require.cache[inquirerPath] = {
+    id: inquirerPath,
+    filename: inquirerPath,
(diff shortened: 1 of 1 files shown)


nemesifier (commented):
Thanks @czarflix, strengthening the `add-org` test so it actually asserts the generated configuration (rather than just that the command runs) is exactly the kind of test hardening I like, and it is self-contained.

A couple of points:

1. Confirm the new `interactiveRunner` subprocess approach is robust in CI (no flaky timing, cleans up any files it writes under `organizations/<slug>/`).
2. Make sure the test fails loudly if the config file is not created (i.e. it would actually catch the regression it is meant to catch); a quick "break the generator, see the test fail" check is worth doing.

Looks good; minor confirmations and it is ready.

PR #1066, opened 2026-03-20
[fix] Unhandled exception when parsing invalid captive portal action url in status.js

### LINKED ISSUE #1065: [bug] Unhandled exception when parsing invalid captive portal action url in status.js
**Describe the bug**
The handlePostMessage function parses captivePortalLoginForm.action using the URL constructor.

If the action value is empty or malformed, new URL() may throw an exception. While this does not always break the UI, it introduces an unhandled runtime error in the message handler.

**Steps To Reproduce**
1. Open the login page (e.g. http://localhost:8080/default/login)
2. Open browser console
3. Modify the form action manually:
   document.querySelector("form").action = "invalid-url"
4. Trigger postMessage handling:
   window.postMessage({ type: "authMessage", message: "test" }, window.location.origin)
5. Observe error from URL constructor in console
**Expected behavior**

**Screenshots**
N/A

**System Informatioon:**

- OS: Fedora Linux 43
- Node.js Version: (22.22.1)
- Browser: Firefox (latest)

Ontiomacer:
Handle invalid or empty captivePortalLoginForm.action values safely by wrapping URL parsing in a try/catch block.
Also ensure safe destructuring of event.data to avoid runtime errors when the message payload is undefined.
This improves robustness of the postMessage handler without changing existing behavior.

## Checklist

Closes #1065.

## Description of Changes
The handlePostMessage function parses captivePortalLoginForm.action using the URL constructor.

If the action value is empty or malformed, URL parsing may throw an exception. This change wraps the parsing logic in a try/catch block and validates the action before parsing.

Additionally, event.data is safely destructured to prevent runtime errors when the message payload is undefined.

These changes improve robustness of the message handler without altering existing behavior.

**No functional changes are introduced.**
## Screenshot
N/A


nemesifier (changes requested):
Thanks @Ontiomacer, this is a good defensive fix. Wrapping `new URL(captivePortalLoginForm.action)` in a try/catch (and guarding with `?.trim()`) prevents a crash when the action is empty or malformed, and `const {...} = event.data || {}` avoids destructuring an undefined payload. The trusted-origin check is preserved correctly: an unparsable action simply yields `actionOrigin = null`, so only the real `window.location.origin` is trusted, which is the safe default.

One ask:

- Add a test for the invalid/empty `action` case (and the `event.data` undefined case) to lock in that a bad URL does not throw and that an untrusted origin is still rejected. The tests box is unchecked.

Good, small, and correct.

Ontiomacer:
Thanks @nemesifier for the review and the detailed feedback.

I’ve added regression tests for:

* invalid/empty `captivePortalLoginForm.action`
* undefined `event.data`
* untrusted origin handling

I also cleaned up the earlier test issues and updated the branch accordingly.
Sorry for the noisy intermediate commit while I was fixing the test logic 
I should have validated the mock behavior more carefully before pushing.
The latest revision should now be consistent with the requested behavior and the test suite is passing locally.

Issue #314, opened 2021-07-23
[change] Eliminate redundancy of header HTML


nemesisdesign:
The header HTML is kind of duplicated, once is a desktop version and one is a mobile version.

I wonder if we can avoid this and handle everything with CSS.

It's kind of painful to customize this part because of this duplication (double amount of work).

Red-0111:
Hi, I would like to work on this issue, please tell me how to proceed :)

codesankalp:
Hi @Red-0111, Please check `client/components/header/`.
In this headers are defined for mobile and desktop separately.
**mobile**:
https://github.com/openwisp/openwisp-wifi-login-pages/blob/f2af7f1940aa6f41d2b2d2b35567c016382396b2/client/components/header/header.js#L145-L148
**desktop**:
https://github.com/openwisp/openwisp-wifi-login-pages/blob/f2af7f1940aa6f41d2b2d2b35567c016382396b2/client/components/header/header.js#L59-L62

In this issue, it is mentioned to get rid of one header using CSS.

Red-0111:
Yes please give me some time I am trying to figure it out.

codesankalp:
> Yes please give me some time I am trying to figure it out.

Great, If you need any help feel free to ask :+1:

iamvaishnav:
Is anyone working on this?

Red-0111:
> Is anyone working on this?

Yes, it would be great to have some help :)

codesankalp:
> > Is anyone working on this?
> 
> Yes, it would be great to have some help :)

What issues are you facing @Red-0111?

Red-0111:
> > > Is anyone working on this?
> > 
> > Yes, it would be great to have some help :)
> 
> What issues are you facing @Red-0111?

While running the headless test for yarn the test suites fail and it shows command failed with exit code 1

codesankalp:
Have you tried the correct way of running browser-test as described in README?

@Red-0111 If you have implemented this, please make a draft pull request. I will review the PR and then can tell more about the error.

Red-0111:
> Have you tried the correct way of running browser-test as described in README?
> 
> @Red-0111 If you have implemented this, please make a draft pull request. I will review the PR and then can tell more about the error.
> 

Yes I followed the same steps as in the README. I was installing dependencies so what should I make the pull request for?

codesankalp:
> I was installing dependencies so what should I make the pull request for?

[...]

PR #1048, opened 2026-03-03
Issues/314 enhanchment title unify header

sanjayahari:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [x] I have manually tested the changes proposed in this pull request.
- [x] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [x] I have updated the documentation.

## Reference to Existing Issue

Closes #<issue-number>.

Please [open a new issue](https://github.com/openwisp/openwisp-wifi-login-pages/issues/new/choose) if there isn't an existing issue yet.

## Description of Changes

Please describe these changes.

## Screenshot

Please include any relevant screenshots.
Screenshot 2026-03-04 004331
Screenshot 202
<img width=
6-03-04 004343" src="https://github.com/user-attachments/assets/b7e0bfb1-30da-4794-9660-5c727daa5d6c" />

PR #1032, opened 2026-02-22
[change] extracted Header logic into reusable methods to remove HTML #314

prathmeshkulkarni-coder:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [x] I have manually tested the changes proposed in this pull request.
- [x] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [ ] I have updated the documentation.

## Reference to Existing Issue

Please [open a new issue](https://github.com/openwisp/openwisp-wifi-login-pages/issues/new/choose) if there isn't an existing issue yet.

## Description of Changes

As part of the modernization effort outlined , I have refactored the Header.js component to remove structural duplication.

Improvements:

DRY Implementation: Extracted rendering logic into renderLogos, renderNavLinks, and renderLanguageButtons.

Reduced JSX Bloat: Eliminated redundant HTML blocks that were previously mirrored for desktop and mobile views.

Maintainability: Standardized how internal/external links are processed across all viewports.

This refactor simplifies the component's structure, making it a cleaner candidate for the React 19 migration.

## Mobile View
image

##  Desktop View
image

PR #1068, opened 2026-03-21
[change]  Fix header duplication #314

kankaa16:
## Checklist

- [] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [] I have updated the documentation.

Related to #314 

## Description of Changes

Reduced duplicated logic between desktop and mobile headers.
No functional changes.

### DIFF
--- a/client/components/header/__snapshots__/header.test.js.snap
+++ b/client/components/header/__snapshots__/header.test.js.snap
@@ -57,8 +57,7 @@ exports[`<Header /> rendering should render sticky msg and close it on clicking 
         className="header-row-2-inner"
       >
         <a
-          className="header-link header-desktop-link
-                    header-link-1 button"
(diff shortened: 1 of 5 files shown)


nemesifier (commented):
Thanks @kankaa16 for working on #314. The approach is reasonable, but there are four open PRs for this same issue (#1080, this, #1048, #1032), and we can only take one.

Rather than review all four in depth, I would like to consolidate on a single canonical PR. #1080 is currently the cleanest and most focused candidate. Please coordinate with the other contributors so we settle on one implementation; if there is something this PR does better than #1080 (a cleaner abstraction, better tests), call it out and we can fold it in. Otherwise I would close this in favor of the canonical one.

Whichever we pick must: preserve header rendering across all org configurations (logo/no-logo, custom links, language switcher), update snapshots intentionally, and not pull in unrelated churn.

PR #1078, opened 2026-03-25
Auto-generated test-translation.json from en.po file

### LINKED ISSUE #902: [feature] Automate generation of client/test-translation.json from .po files
**Is your feature request related to a problem? Please describe.**
The `client/test-translation.json` file currently requires manual updates to add or change translations for strings. This is repetitive work since the translations already exist in the corresponding `.po` files. If one misses updating test-translation.json file, then the snapshots created for tests only include the `msgid`. This creates inconsistency and increases maintenance overhead. 

**Describe the solution you'd like**
A script or tool that can automatically generate or update `client/test-translation.json` by extracting translations directly from the `.po` files.

 A possible solution could be to add this process to the `yarn test` command, so the file is populated automatically before tests are run.

Kasturijadhav14:
The test-translation.json file was being updated manually
which was repetitive work and easy to forget. If someone
missed updating it, the tests would break.

I wrote a small script that reads the en.po file and
automatically creates the test-translation.json file.
I also added it to the yarn setup command so it runs
on its own every time.

Changes made:

Added config/build-test-translation.js
Updated yarn setup in package.json to run the new script
Fixes https://github.com/openwisp/openwisp-wifi-login-pages/issues/902


nemesifier:
Thank you for your contribution. We've seen a high volume of PRs recently with dubious quality, and unfortunately we don't have the bandwidth to review and iterate on each one thoroughly. We're closing this PR to keep the queue manageable. We appreciate your understanding.

PR #1081, opened 2026-03-28
Refactor: Extracted session utility functions from status component

JarrarHaiderNemati:
@nemesifier 

Extracted getDuration, getDateTimeFormat and getUserCheckFormattedValue from the Status component into a separate client/utils/session-utils.js utility file.

These are pure functions with no dependency on component state or lifecycle. Extracting them is a first step toward addressing #918 : separating concerns in the Status component to improve maintainability and testability. Each function can now be tested and reused independently.

Tests: All 91 status tests pass.

Am i on the right track ?

### DIFF
--- a/client/utils/session-utils.js
+++ b/client/utils/session-utils.js
@@ -14,7 +14,9 @@ export const getDuration = (seconds) => {
 
 export const getDateTimeFormat = (language, time_option, date) => {
   if (typeof Intl !== "undefined") {
-    return new Intl.DateTimeFormat(language, time_option).format(new Date(date));
+    return new Intl.DateTimeFormat(language, time_option).format(
+      new Date(date),
+    );
   }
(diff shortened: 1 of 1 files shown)


nemesifier (commented):
Thanks @JarrarHaiderNemati, extracting session helpers out of the Status component is a reasonable cleanup.

The main issue is overlap: #1087 also extracts session-related code (the sessions table) from Status. The two touch the same area and should be coordinated so they do not conflict. Let's decide on one coherent extraction plan: if these are complementary (utils here, table in #1087), sequence them and make sure the table PR consumes these utils; if they overlap, consolidate into one.

Also, since this only adds functions without removing the originals from Status, please confirm Status is actually updated to use the extracted utilities (otherwise it is dead duplication), and add or move tests for the extracted functions.

PR #1088, opened 2026-03-31
[fix] Harden email validation and typo detection #1044

### LINKED ISSUE #1044: [bug] Weak Email Validation Pattern in Registration and Login Forms
**Describe the bug**
The email validation logic in the login and registration forms is too permissive. It currently uses a loose regex pattern (.+@.+\..+) defined in the organization configuration, which fails to catch invalid TLDs or common domain typos. For example, the system accepts user@gmal.com as a valid email address without any warning or rejection.

**Steps To Reproduce**

1. Go to the Registration pageor Login page.
2. Enter an email address with a common domain typo, such as vikramk2101@gmal.com.
3. Observe that the form accepts the input as valid (the browser's built-in validation passes and no custom error is       triggered).
4. See that the invalid email is successfully submitted or displayed in the user's status after "login" as shown in the          screenshot.

**Expected behavior**
The email field should implement a more robust validation pattern that:

Enforces a standard format with a valid TLD (e.g., [a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$).
(Bonus/Enhancement) Detects common domain typos (like gmal.com instead of gmail.com) and warns the user to prevent account recovery issues later.

**Screenshots**

Image

**System Informatioon:**

OS: Windows (WSL Ubuntu)
Node.js: v18
Browser: Chromium-based (Chrome/Edge)

AkramZaabi:
## Checklist

Closes #1044.

## Description of Changes
This PR fixes weak email validation in login and registration flows.

**What was happening**

The previous email pattern was too permissive (`.+@.+\..+`) and accepted invalid/typo-prone emails like `vikramk2101@gmal.com` without warning.

**What changed**

- Introduced shared email validation utilities to:
  - enforce a safer default email regex when legacy permissive patterns are detected;
  - detect common domain typos and suggest corrections (e.g. `gmal.com` → `gmail.com`).
- Applied stronger effective email validation in:
  - login username/email input (when configured as `type: email`);
  - registration email input.
- Added pre-submit typo checks in both login and registration to prevent accidental submission with known domain typos.
- Updated default configuration and generator templates to use a stronger email regex.
- Added/updated tests and snapshots for the new email validation utility, login flow, and registration flow.

**Validation performed**

- Targeted Jest suites for login, registration, and email validation utility passed.
- Dependency usage check passed.

## Screenshot
image_2026-03-31_124450364


nemesifier (changes requested):
Thanks @AkramZaabi, email typo detection (catching "gmial.com" etc.) for #1044 is a nice UX improvement and the test coverage is good.

The one thing I need verified before merge is the username field pattern change:

- You changed the field pattern from `.+@.+\..+` to `[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$`. That username field accepts an email or a username or a phone number (per its own title text). An email-shaped pattern that requires an `@` and a TLD will reject plain usernames and phone numbers via HTML5 validation. The previous pattern also required `@`, so this may be a pre-existing limitation rather than a regression, but please confirm: does this field ever need to accept non-email logins? If so, the stricter pattern must only apply when the input is actually an email (e.g. set dynamically), not unconditionally. If the field is email-only in practice, say so in the PR.
- Minor: HTML5 `pattern` is implicitly anchored at both ends, so the trailing `$` is redundant (harmless).

Please confirm the username/phone case and, if needed, scope the email pattern to email input only. The typo-detection part looks good.

PR #1104, opened 2026-04-21
[feature] Added support for cross-organization login #1055

### LINKED ISSUE #1055: [feature] Allow users joining different org with different rules
⚠️ **Not suited to beginner contributors!**

Depends on https://github.com/openwisp/openwisp-radius/issues/692.

It must be possible for a user to join a different org with different identity verification requirements or pricing without breaking security.

pandafy:
## Checklist

Closes #1055 

## Description of  changes 

Adds a dedicated cross-organization flow component to centralize the
post-authentication onboarding logic for users logging in to a different
organization.

The new flow ensures organization-specific requirements are completed
before granting access in the target organization:

- users are redirected to mobile phone verification when phone
  verification is enabled in the target organization;

- users are redirected to the complete-signup page when subscriptions
  are enabled, allowing them to review and switch subscription plans
  before completing the onboarding process.
  

## Blockers 

- [ ] https://github.com/openwisp/openwisp-radius/pull/698


.github/workflows/ci.yml
```diff
@@ -107,7 +107,8 @@ jobs:
 
       - name: Get openwisp-radius
         run: |
-          git clone --depth=1 https://github.com/openwisp/openwisp-radius/ openwisp-radius
+          # TODO: Remove branch before merging
+          git clone --depth=1 --branch issues/692-different-identity-verification https://github.com/openwisp/openwisp-radius/ openwisp-radius
```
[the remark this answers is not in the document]
pandafy: @coderabbitai we will remove this before merging!

.github/workflows/ci.yml
```diff
@@ -131,6 +132,7 @@ jobs:
           pip install -U pip setuptools wheel
           pip install -U -e ".[saml]"
           pip install -U -r requirements-test.txt
+          pip install --no-deps --no-cache-dir --force-reinstall "https://github.com/openwisp/openwisp-users/tarball/issues/497-export-users"
```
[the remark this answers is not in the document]
pandafy: @coderabbitai we will remove this before merging!

pandafy:
@coderabbitai do a full review of the changes, focus on security, bugs, performance, unused code or outdated comments/docs. Ensure that the code follows OpenWISP Conventions.  Flag if any user facing string is untranslated.

nemesifier:
@coderabbit full review

nemesifier (changes requested):
## Blockers

**CI workflow still pinned to unmerged upstream branches.** `.github/workflows/ci.yml:111` clones `openwisp-radius` from `issues/692-different-identity-verification` and `:135` force-reinstalls `openwisp-users` from `issues/497-export-users`. The PR description still lists `openwisp/openwisp-radius#698` as a blocker. Once those land, the workflow needs to revert to the default clone and the released `openwisp-users` before this branch can merge — otherwise master CI is left pinned to mutable feature branches.

**Untranslated user-facing string `REGISTRATION_TITLE`.** `client/components/complete-signup/complete-signup.js:60` and `:297` call `` t`REGISTRATION_TITLE` ``, but no `.po` file (en/de/es/fur/it/ru/sl) defines this msgid. The existing convention is the suffix `_TITL` (see `REGISTRATION_TITL` in `client/components/registration/registration.js:75` → "Sign up" in `i18n/en.po:349`). Result: the heading and `<title>` will render the literal "REGISTRATION_TITLE" in every locale. Either rename the key to `REGISTRATION_TITL` (reusing the existing translation) or add new entries to every catalog and re-extract.

## Bugs

**`server/utils/logger.test.js` does not exercise the new redaction logic.** The top-level `jest.mock("../utils/logger", () => ({logResponseError: jest.fn()}))` resolves to the file under test (from `server/utils/`, `../utils/logger` → `server/utils/logger`). After `jest.resetModules()` + `require("./logger")` the test still receives the auto-mocked `jest.fn()` — verified locally: `jest.isMockFunction(logResponseError)` returns `true` inside the test, and the test names themselves admit "sanitization happens internally" / "sanitization is internal". Every assertion is the trivial `expect(jest.fn()).toHaveBeenCalledWith(error)` and `redactHeaders` / `sanitizeResponse` / `sanitizeRequest` in `server/utils/logger.js:122-151` are never run. Drop the top-level mock, spy on the underlying `Logger.error` / `Logger.info`, and assert the payload no longer contains the bearer token.

**Duplicate toast on plans-fetch failure.** `client/utils/get-plans.js:26` calls `toast.error(t\`ERR_OCCUR\`)` and then invokes `failureCallback`, and `client/components/complete-signup/complete-signup.js:100` (`handlePlansFailure`) calls `toast.error(t\`PLANS_FETCH_ERR\`)` as well — so a single failure stacks two toasts ("Error occurred!" then "Failed to load plans"). Either gate the toast in `get-plans.js` behind the absence of a callback, or drop the toast in `handlePlansFailure` (the inline alert + retry button is already shown on the page).

## Security (defense-in-depth)

**Proxy unconditionally allows `method: ""`.** `server/controllers/registration-method-controller.js:23` initializes `allowedMethods = [""]`, so even when the org has `mobile_phone_verification: true` and/or `subscriptions: true`, a `pending_verification` user can POST `{"method": ""}` directly to `/api/v1/{org}/account/registration-method` and the proxy forwards it. The frontend only ever sends `""` when *both* settings are disabled (`complete-signup.js:139-145`), so the empty string is the org-internal "no verification required" marker; the proxy validation should mirror that:

```js
const allowedMethods = [];
if (!settings.subscriptions && !settings.mobile_phone_verification) allowedMethods.push("");
if (settings.subscriptions) allowedMethods.push("bank_card");
if (settings.mobile_phone_verification) allowedMethods.push("mobile_phone");
```

The backend should also enforce this, but the issue explicitly says "without breaking security" so it is worth tightening here too.

pandafy:
> ## Security (defense-in-depth)
> **Proxy unconditionally allows `method: ""`.** `server/controllers/registration-method-controller.js:23` initializes `allowedMethods = [""]`, so even when the org has `mobile_phone_verification: true` and/or `subscriptions: true`, a `pending_verification` user can POST `{"method": ""}` directly to `/api/v1/{org}/account/registration-method` and the proxy forwards it. The frontend only ever sends `""` when _both_ settings are disabled (`complete-signup.js:139-145`), so the empty string is the org-internal "no verification required" marker; the proxy validation should mirror that:
> 
> ```js
> const allowedMethods = [];
> if (!settings.subscriptions && !settings.mobile_phone_verification) allowedMethods.push("");
> if (settings.subscriptions) allowedMethods.push("bank_card");
> if (settings.mobile_phone_verification) allowedMethods.push("mobile_phone");
> ```
> 
> The backend should also enforce this, but the issue explicitly says "without breaking security" so it is worth tightening here too.

An organization can have a registration option which does not require verification when when mobille_phone_verification and subscription is set to `true.

nemesifier (changes requested):
Thanks for working on this. I found a couple of small issues that should be addressed before merging.

## 1. Inline errors are stored under the wrong key

In `client/components/complete-signup/complete-signup.js`, the new flow stores some errors under `nonfield`, but the UI reads them from `nonField` through `getError(errors)`.

As a result, if `updateRegistrationMethod` or `upgradePlan` fails, the toast is shown, but the inline error area does not render the error message.

**Suggested fix:** use the existing `nonField` key consistently when setting these errors, so it matches the rest of the component/error handling convention.

## 2. `get_phone_token.py` can crash before using its fallback phone number

In `browser-test/get_phone_token.py`, the script inserts `"browser-test"` into `sys.argv`, then reads `sys.argv[2]` directly.

If the script is run without a phone number argument, this raises `IndexError`, so the fallback to `testData["mobileVerificationTestUser"]["phoneNumber"]` is never reached.

**Suggested fix:** check that the phone number argument exists before reading it. For example, after inserting `"browser-test"`, use a guard like `len(sys.argv) > 2` before accessing `sys.argv[2]`; otherwise use the existing fallback value.

.github/workflows/ci.yml
nemesifier: we can remove this now

client/components/organization-wrapper/organization-wrapper.js
```diff
@@ -217,26 +227,44 @@ export default class OrganizationWrapper extends React.Component {
...
+                            <Navigate to={`/${orgSlug}/complete-signup`} />
+                          );
+                        }
+                        if (needsVerifyPhone) {
+                          return (
+                            <Navigate
+                              to={`/${orgSlug}/mobile-phone-verification`}
+                            />
+                          );
+                        }
```
nemesifier, line 267: @pandafy are you aware of these nested if statements here?

client/components/organization-wrapper/organization-wrapper.js
```diff
@@ -312,6 +340,32 @@ export default class OrganizationWrapper extends React.Component {
...
+                        if (needsVerifyBankCard) {
+                          return <Navigate to={`/${orgSlug}/payment/draft`} />;
+                        }
+                        return <Navigate to={`/${orgSlug}/status`} />;
+                      }
```
nemesifier, line 361: same here

nemesifier (changes requested):
A new component is being added but there's no mention of this in the PR description.

Can you please expand the PR description to explain how the flow is changed? I need this for the change log.

client/components/organization-wrapper/organization-wrapper.js
pandafy: @nemesifier do you prefer evaluating `isAuthenticated` twice for each branch?

pandafy: @nemesifier these changes are intentional.

.github/workflows/ci.yml
```diff
@@ -107,7 +107,6 @@ jobs:
 
       - name: Get openwisp-radius
         run: |
-          git clone --depth=1 https://github.com/openwisp/openwisp-radius/ openwisp-radius
```
nemesifier, line 110: Ok I didn't mean to remove the entire line only the custom branch I guess

client/components/complete-signup/complete-signup.js
```diff
@@ -0,0 +1,362 @@
...
+
+  handlePlanChange = (event) => {
+    const planIndex = event.target.value;
+    const {plans} = this.state;
+    const selectedPlan = plans[planIndex];
+    if (!selectedPlan) {
+      return;
+    }
+    this.setStateSafe({selectedPlan: planIndex});
+    this.handleSubmitPlan(planIndex);
```
nemesifier, line 274: **Medium:** `handlePlanChange` fires `handleSubmitPlan` on radio `onChange`. After a submission failure, `selectedPlan` stays set; clicking the already-checked radio does not fire `onChange` again. Single-plan orgs (or retrying the same plan) are stuck with no retry mechanism.

**Fix:** Reset `selectedPlan` to `null` in the error handler at ~line 250 so the radio can be re-selected, or add a Retry button in the plan-select area.

nemesifier (changes requested):

nemesifier (approved):

PR #1146, opened 2026-07-16
[fix] Show payment failure page for plan-upgrade payments #1145

### LINKED ISSUE #1145: [bug] Failed plan-upgrade payment redirects to /status instead of /payment/failed
**Describe the bug**
A user who registered via the `mobile_phone` flow (and is already verified) gets silently
redirected from `/{orgSlug}/payment/failed` to `/{orgSlug}/status` when a plan-upgrade payment
fails, instead of seeing an explanation with a way to retry the payment or go back to their
account.

**Steps To Reproduce**

1. Register with the `mobile_phone` verification flow and get verified.
2. Exhaust the plan quota so the status page offers a plan upgrade.
3. Choose a paid plan upgrade and proceed to the payment gateway.
4. Make the payment fail (or cancel it) at the gateway.
5. Get redirected back to `/{orgSlug}/payment/failed`.
6. Observe: you are immediately bounced to `/{orgSlug}/status` with no explanation of what
   happened to the payment.

**Expected behavior**
`/payment/failed` should render, telling the user the payment failed, with two options: retry
the payment, or go back to their account without being logged out (their existing paid plan is
still active).

**Root cause**
`PaymentStatus.render()` (`client/components/payment-status/payment-status.js`) has two guards
written for the `bank_card` registration flow:

- `method && method !== "bank_card"`
- `["failed", "draft"].includes(status) && isVerified === true`

A plan upgrade never changes `method` or `is_verified`; those only flip once a payment is
confirmed. So a `mobile_phone` user mid-upgrade still carries `method: "mobile_phone"` and
`is_verified: true` through the whole payment window and trips both guards. There is currently no
signal in the user data that distinguishes "verified, unrelated to any payment" from "verified,
but mid a failed upgrade payment."

**Screenshots**
N/A

**System Informatioon:**

- OS: any
- Node.js Version: any
- Browser and Browser Version (if applicable): any (logic bug, not browser-specific)

### LINKED ISSUE #1147: [bug] Plan upgrade skips captive-portal login before payment when payment_requires_internet is true
⚠️ **Not suitable for beginners**

**Describe the bug**

When an organization has `payment_requires_internet = true`, a user who upgrades their plan is
sent straight to the payment gateway without first being logged into the captive portal. The
payment gateway lives outside the walled garden, so a user whose quota is exhausted has no
internet access and cannot reach it. The upgrade flow never establishes the internet session that
`payment_requires_internet` is meant to guarantee.

This affects the plan-upgrade flow specifically. The bank_card registration/draft flow handles
`payment_requires_internet` correctly, but the upgrade flow was never wired into the same
mechanism.

**Steps To Reproduce**

1. Configure an organization with `payment_requires_internet = true` and a payment gateway that is
   outside the captive-portal walled garden.
2. Register and verify a user with `mobile_phone`, on a free plan.
3. Consume the free plan quota until it is exhausted (no internet).
4. On `/status`, open the upgrade modal and select a paid plan.
5. Observe that the app navigates directly to `/payment/process` (and then to the external
   gateway) instead of first logging the user into the captive portal. The gateway is unreachable.

**Expected behavior**

When `payment_requires_internet` is true, the upgrade flow should log the user into the captive
portal first (establishing payment-window internet), then send them to `/payment/process`,
mirroring the bank_card draft flow.

**Technical notes**

- `upgradeUserPlan` (`client/components/status/status.js:459-502`) ignores
  `payment_requires_internet` and calls `navigate('/payment/process')` directly. It does not set
  `proceedToPayment` and does not route through `/status` the way `paymentProceedHandler` does
  (`client/components/payment-status/payment-status.js:139-152`).
- The login-first bridge in `finalOperations` (`client/components/status/status.js:273-285`) only
  fires for `needsVerify("bank_card", ...)`, which is false for a mobile_phone upgrader
  (`method === "mobile_phone"`, `is_verified === true`), so the upgrade case is never handled even
  if `proceedToPayment` were set.

**System information**

- Config-dependent bug; requires `payment_requires_internet = true`.

pandafy:
## Checklist

Closes #1145
Fixes #1147

## Description of Changes

A mobile_phone-registered, already-verified user whose plan-upgrade payment failed was silently redirected to /status instead of /payment/failed, because the redirect guards in PaymentStatus were written only for the bank_card registration flow and could not tell a mid-upgrade failure apart from an unrelated already-verified visit.

Read the new `in_upgrade` flag to recognize this case and let the user retry the payment or go back to their account without being logged out, since their existing plan stays active either way.

### DIFF
--- a/client/components/payment-status/payment-status.js
+++ b/client/components/payment-status/payment-status.js
@@ -10,6 +10,8 @@ import Loader from "../../utils/loader";
 import Contact from "../contact-box";
 import validateToken from "../../utils/validate-token";
 import handleLogout from "../../utils/handle-logout";
+import cancelUpgradePlan from "../../utils/cancel-upgrade-plan";
+import logError from "../../utils/log-error";
 
 export default class PaymentStatus extends React.Component {
   constructor(props) {
@@ -65,10 +67,39 @@ export default class PaymentStatus extends React.Component {
     }
   }
 
-  // a user who gives up on a plan upgrade keeps the plan they already
-  // paid for, so they are sent back to the status page without logging out
-  backToStatus = () => {
-    const {orgSlug, navigate} = this.props;
+  // a user who gives up on a plan upgrade keeps their existing plan,
+  // so they are sent back to the status page without logging out.
+  // The backend must cancel the pending upgrade order and stale
+  // upgrade/login flags must be cleared, otherwise status.js can still
+  // bounce the user back to payment/draft.
+  backToStatus = async () => {
+    const {orgSlug, navigate, setUserData, userData, language} = this.props;
+    try {
+      const response = await cancelUpgradePlan(
+        orgSlug,
+        userData.auth_token || userData.key,
+        language,
+      );
+      setUserData({
+        ...userData,
+        ...response,
+        auth_token: response.auth_token || response.key || userData.auth_token,
+        proceedToPayment: false,
+        mustLogin: undefined,
+      });
+    } catch (error) {
+      if (!error.response || error.response.status !== 404) {
+        toast.error(t`ERR_OCCUR`);
+        logError(error, "Error while cancelling plan upgrade");
+      }
+      setUserData({
+        ...userData,
+        in_upgrade: false,
+        proceedToPayment: false,
+        payment_url: null,
+        mustLogin: undefined,
+      });
+    }
     navigate(`/${orgSlug}/status`);
   };
 
--- a/client/components/payment-status/payment-status.js
+++ b/client/components/payment-status/payment-status.js
@@ -84,6 +84,7 @@ export default class PaymentStatus extends React.Component {
         ...userData,
         ...response,
         auth_token: response.auth_token || response.key || userData.auth_token,
+        in_upgrade: false,
         proceedToPayment: false,
         mustLogin: undefined,
         mustLogout: true,
@@ -151,7 +152,6 @@ export default class PaymentStatus extends React.Component {
     }
 
     // draft case
-    // if (isAuthenticated === false && status === "draft") {
     if (status === "draft") {
       return this.renderDraft(isDraftUpgrade);
     }
@@ -162,7 +162,9 @@ export default class PaymentStatus extends React.Component {
       return redirectToStatus();
     }
 
-    // Show loader while validateToken resolves in_upgrade state on cold reload
+    // Prevent a brief render of the non-upgrade failed page while validateToken
+    // is resolving on a cold reload. The draft flow doesn't need this because it
+    // doesn't branch on in_upgrade until the payment-proceed step.
     if (status === "failed" && isTokenValid === null) {
       return <Loader />;
     }
@@ -255,7 +257,7 @@ export default class PaymentStatus extends React.Component {
   }
 
   renderFailed(isFailedUpgrade = false) {
-    const {orgSlug, userData} = this.props;
+    const {orgSlug, userData, settings} = this.props;
     // User might have exhausted the temporary session quota while completing
     // the payment. Thus, we need to login them back in the captive portal
     // before directing them to the payment gateway.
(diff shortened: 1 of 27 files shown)


pandafy:
@coderabbitai Re: **"Run the token-loading guard before user-data-dependent redirects."** (`payment-status.js:100-113`, also applies to 129-136)

I don't think this is reachable in the current code. `method`, `is_verified`, and `in_upgrade` are never set independently: they all come from a single `setUserData(response.data)` call inside `validateToken` (`client/utils/validate-token.js`), and the reducer does a full replacement of `userData` (`client/reducers/organization.js:54-59`), not a partial merge. So there's no render where `method` reflects the upgrade case while `in_upgrade` is still stale.

Traced both reachable states on a cold load:
- Before `setUserData` fires: `userData` is fully empty (`initialState`), so `method` is `undefined`, and the `method && method !== "bank_card"` guard can't fire regardless of `in_upgrade`. Falls through to the loader guard.
- After `setUserData` fires but before `isTokenValid` updates: `method` and `in_upgrade` are already consistent (both reflect the real upgrade), so `isFailedUpgrade` is already `true` and the bank_card guard is correctly suppressed.

The existing test at `payment-status.test.js:115-130` covers the actual reachable case (`userData: {}`) and passes. I couldn't construct a path that produces the split state (`method` set, `in_upgrade` unset) this comment assumes.

Given that, and the follow-up note on the neighboring test in this same review ("Likely an incorrect or invalid review comment"), I'll leave this as-is unless there's a concrete repro.

pandafy:
@coderabbitai full review!

nemesifier (changes requested):
It seems that failed upgrade payments cannot be retried in the iframe flow.

client/components/payment-status/payment-status.js
```diff
@@ -182,8 +198,14 @@ export default class PaymentStatus extends React.Component {
...
-  renderFailed() {
-    const {orgSlug} = this.props;
+  renderFailed(isFailedUpgrade = false) {
+    const {orgSlug, userData} = this.props;
+    const retryUrl = isFailedUpgrade
+      ? `/${orgSlug}/payment/process`
+      : `/${orgSlug}/payment/draft`;
+    // an upgrade which ran out of allowed payment attempts has no
+    // payment URL left to send the user to
+    const showRetry = !isFailedUpgrade || Boolean(userData.payment_url);
```
nemesifier, line 273: Line 206 treats a missing `userData.payment_url` as the terminal case where the retry button must be hidden. That is not the only way this state is reached. In the iframe flow, `client/utils/get-payment-status.js:63` clears `payment_url` for every failed payment before navigating to `/payment/failed`. For an already verified `mobile_phone` upgrader, `client/utils/validate-token.js:25` then skips the token validation request because `radius_user_token` is already present and the user is not an unverified `bank_card` registration, so the new retry URL exposed by the backend is never fetched. The result is a failed upgrade page with only the go-back button, even though the issue asks for a retry option.

The smallest fix I would request is to make the validation path refetch user data when `userData.in_upgrade === true` and `payment_url` is missing, or otherwise preserve/recover the upgrade retry URL before rendering this page. Please add a regression test covering the iframe path where `getPaymentStatusRedirectUrl` clears `payment_url`, `in_upgrade` remains true, and `/payment/failed` still gets a retry URL instead of hiding the retry button.

pandafy:
@coderabbitai full review!

client/components/payment-status/payment-status.js
```diff
@@ -117,30 +162,59 @@ export default class PaymentStatus extends React.Component {
...
         ...userData,
+        mustLogin: true,
         proceedToPayment: true,
       });
+      // Persist so it survives page reload during sync captive portal auth
+      storeValue(captivePortalSyncAuth, `${orgSlug}_mustLogin`, true, cookies);
+      storeValue(
+        captivePortalSyncAuth,
+        `${orgSlug}_proceedToPayment`,
+        "true",
```
pandafy: Why is this storing a string literal?

server/controllers/user-cancel-upgrade-plan-controller.js
```diff
@@ -0,0 +1,59 @@
...
+      );
+      const timeout = conf.timeout * 1000;
+      axios({
+        method: "post",
+        headers: {
+          "content-type": "application/json",
+          Authorization: req.headers.authorization,
+          "accept-language": req.headers["accept-language"],
+        },
+        url: `${host}${userUpgradePlanUrl}/cancel/`,
```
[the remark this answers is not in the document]
pandafy: @coderabbitai this has been added in the OpenWISP backend.

pandafy:
@coderabbitai full review!

pandafy:
@coderabbitai full review!

PR #1167, opened 2026-08-02
[fix] Allow logout from change-password page on expired password #1157

### LINKED ISSUE #1157: [bug] Allow logout on expired-password change page
**Describe the bug**
When a user's password has expired, they are redirected to the change password page. On that page, the user cannot log out from the application until the password is changed, which leaves the user stuck on the password change flow.

Additionally, there is no clearly visible UI message indicating that the password has expired apart from the notification toast. The toast is easy to miss, disappears quickly, and users may not always read it.

**Steps To Reproduce**
1. Sign in with an account whose password has expired.
2. Observe that the user is redirected to the change password page.
3. Try to log out from the application.
4. Notice that logout is not available until the password is changed.

**Expected behavior**
The user should be able to log out even when their password has expired and they are on the change password page.

The UI should also present a more visible and persistent message that clearly informs the user that their password has expired.

**Screenshots**

Image

prathmeshkulkarni-coder:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).

Closes #1157 

## Description of Changes

Previously, when a user's password was expired, they were redirected
to the change-password page with no way to log out — trapping them
in the password change flow.
This patch adds:
- A visible "Password expired" warning banner on the change-password page
- A "Logout" button (replacing the "Cancel" link) when `password_expired === true`,
  allowing users to log out instead of being stuck

## Screenshot

Screenshot from 2026-08-02 12-18-07


prathmeshkulkarni-coder:
@pandafy i added logout button .

i have some suggestions here 
image
We should capitalize this placeholder to be consistent

nemesifier (changes requested):
Why are you  making changes to package.json and yarn.lock? Seem unrelated.

prathmeshkulkarni-coder:
@nemesifier , i run yarn install and setup so it must added newer version and it got commit to the PR .

client/components/status/status.js
```diff
@@ -165,6 +165,20 @@ export default class Status extends React.Component {
         cookies,
       );
       ({userData} = this.props);
+
+      if (mustLogout) {
+        if (captivePortalSyncAuth) {
+          // In synchronous captive portal authentication, the page reloads
+          // after form submission, so handleLogoutIframe() must be called manually here.
+          // (handleLogout() is already triggered when the user clicks the "Logout" button.)
+          this.setState({loggedOut: mustLogout});
```
nemesifier: `setState` is asynchronous. In this synchronous captive-portal branch, `handleLogoutIframe()` runs while `loggedOut` is still false. Here `mustLogout` came from Redux after password change, so there is no cookie or local-storage fallback; the handler returns without calling `logout` and nothing retries it. Please invoke it from the `setState` callback and add a regression test with synchronous auth, `mustLogout: true`, and no stored flag.

client/components/organization-wrapper/organization-wrapper.js
```diff
@@ -283,8 +283,14 @@ export default class OrganizationWrapper extends React.Component {
                   <Route
                     path="logout"
                     element={(() => {
-                      if (isAuthenticated)
+                      if (isAuthenticated && !userData.password_expired)
                         return <Navigate to={`/${orgSlug}/status`} />;
+                      if (isAuthenticated && userData.password_expired)
```
nemesifier: Please remove this entire branch. It is not part of this fix: the password-change button calls `handleLogout()`, sets `mustLogout`, and navigates to `status`; it never navigates here. This only renders `Logout`, which does not dispatch `logout`, clear authentication, or set `isAuthenticated` to false. An authenticated expired-password user who visits this route remains logged in, and "Login again" leads through `status` back to `change-password`.

nemesifier (changes requested):
This PR addresses a legitimate usability issue, but it also includes a bogus authenticated logout route that is neither used by the proposed flow nor capable of logging the user out. Contributors need to trace the existing flow and remove unnecessary changes before requesting review. I have raised this expectation with you several times already, so please take the time to vet the full change carefully before submitting further work.

prathmeshkulkarni-coder:
@nemesifier , I understood  wrapper code was not needed , i mess up because i thought logout.js contain logout code . I forgot that status.js contain all function .  
aside from that i already done all manual testing

prathmeshkulkarni-coder:
@coderabbit review

client/components/status/status.test.js
```diff
@@ -945,6 +945,33 @@ describe("<Status /> interactions", () => {
...
+    const setLoading = jest.fn();
+    wrapper = shallow(<Status {...props} />, {
+      context: {setLoading},
+    });
+    const status = wrapper.instance();
+    status.logoutIframeRef.current = {};
+    const handleLogoutIframe = jest.spyOn(status, "handleLogoutIframe");
+
+    await tick();
+
```
nemesifier: avoid needless blank lines

nemesifier (changes requested):
Use blank lines sparingly.

client/components/status/status.test.js
prathmeshkulkarni-coder: Done

client/components/status/status.js
```diff
@@ -165,6 +165,21 @@ export default class Status extends React.Component {
         cookies,
       );
       ({userData} = this.props);
+
+      if (mustLogout) {
+        if (captivePortalSyncAuth) {
+          // In synchronous captive portal authentication, the page reloads
+          // after form submission, so handleLogoutIframe() must be called manually here.
+          // (handleLogout() is already triggered when the user clicks the "Logout" button.)
+          this.setState({loggedOut: mustLogout}, () => {
```
nemesifier, line 178: `handleLogout()` in this comment is the shared utility called by `PasswordChange`, not `Status.handleLogout()`. The utility only sets Redux `mustLogout` and navigates here; it does not query the active session or submit `#cp-logout-form`. With synchronous auth and no stored marker, this branch immediately clears WLP authentication through `handleLogoutIframe()`, leaving an active captive-portal session open. I would call `this.handleLogout(false, repeatLogin)` on the initial Redux-only transition, and reserve this completion path for the reload after the `_self` logout form has persisted `${orgSlug}_mustLogout`.

nemesifier (changes requested):
The synchronous captive-portal path still bypasses the actual logout form submission.

prathmeshkulkarni-coder:
@nemesifier changes are done CI build is running for 20min .

PR #1172, opened 2026-08-12
[fix] Fixed misleading password reset error messages

pandafy:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).

## Description of Changes

Users saw raw backend field errors (e.g. "token: Invalid value") that gave no actionable information, and weak/invalid password errors from the backend were silently dropped from the toast.

Invalid uid/token errors (including a malformed uid, which the backend reports under non_field_errors) now show a generic "Invalid password reset URL" message instead, while password validation errors from the backend are now shown to the user. The full error is still logged to the console for debugging.

### DIFF
--- a/browser-test/password-confirm.test.js
+++ b/browser-test/password-confirm.test.js
@@ -20,7 +20,7 @@ describe("Selenium tests for <PasswordReset />", () => {
     await tearDown(driver);
   });
 
-  it("should show not found on password confirm", async () => {
+  it("should show invalid password reset URL error on password confirm", async () => {
     await driver.get(urls.passwordConfirm("uid", "token"));
     const data = initialData();
     const password = await getElementByCss(driver, "input#password");
@@ -34,6 +34,8 @@ describe("Selenium tests for <PasswordReset />", () => {
     submitBtn.click();
     const successToastDiv = await getElementByCss(driver, "div[role=alert]");
     await driver.wait(until.elementIsVisible(successToastDiv));
-    expect(await successToastDiv.getText()).toEqual("Not found.");
+    expect(await successToastDiv.getText()).toEqual(
+      "Invalid password reset URL",
+    );
   });
 });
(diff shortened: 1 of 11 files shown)


pandafy:
### Why WLP CI started failing

https://github.com/openwisp/openwisp-radius/pull/757 removed the custom `validate_user()` override from `PasswordResetConfirmView`.

Previously, that override decoded the `uid` itself and raised `Http404` for an invalid or unknown user, so the confirm endpoint returned:

`404 {"detail": "Not found."}`

The endpoint is now delegated to openwisp-users, which reuses the dj-rest-auth serializer. Invalid `uid`/token pairs are therefore returned as normal DRF validation errors:

* malformed `uid` → `400 {"non_field_errors": ["'x' is not a valid UUID."]}`
* unknown `uid` → `400 {"uid": ["Invalid value"]}`
* bad token → `400 {"token": ["Invalid value"]}` (unchanged)

The WLP password-confirm component was relying on the old response shape. It first looked for `detail` and then accessed `token[0]`. With the new responses, neither key is necessarily present, causing a `TypeError`. The loader then remained stuck and no error message was shown.

This PR updates the component to handle the new error keys (`new_password1`, `new_password2`, `uid`, `token`, and `non_field_errors`). For `uid`/`token` errors, it shows a single user-friendly **"Invalid password reset URL"** message, since there is nothing the user can fix on the form, while keeping the original backend error in the console for debugging.

nemesifier:
@coderabbitai any outstanding issue?

nemesifier:
@coderabbitai resolve

nemesifier (approved):

